PR #12, opened 2019-02-14
Testing two shortcut styled referenced links

jlento:


penikune:
I reject this on request by Juha Lento.

PR #15, opened 2019-04-11
Proposal for Github Pull Request workflow

martbhell:
- Also added "git commit -v" in places as I found that very helpful to
   see which file my git commit is actually modifying


attesillanpaa:
Useful additional info

PR #22, opened 2019-05-13
Updated Pouta hierarchy

oscarkraemer:
- Updated css to support additional hierarchy
- Merged some object-storage pages.
- Added FAQ (https://research.csc.fi/pouta-faq) it is still
incomplete
- Added .gitignore


attesillanpaa:
Do you think we should have a separate FAQ for every topic or one for all? We kind of agreed to add a "knowledge base" or FAQ to the main hierarchy level. It could have subtopics then.

attesillanpaa:
Tried to push two commits to fix links, but didn't seem to succed. FAQ can be edited later.

PR #23, opened 2019-05-13
Added simple travis testing.

oscarkraemer:
- Added travis testing.
- Added test for typos. At the moment it does not fail if there are less than 2500 typos.
- Added a dictionary for accepted words: Taito, Pouta, Puhti, OpenStack .a.s.o
- Added test for MkDocs WARNINGS. This will basically fail if links are not working.
- Changed a taito -> Taito.


oscarkraemer:
This should probably be squash-merged.

Nortamo:
Looks good. 
I think that the test for typos will work if people actively add exceptions. (This could be in the instructions)

Currently we don't test for a successful build: So if `mkdocs build` fails all the test pass. 
I would suggest that we add a test for this and the do a squash merge.

If it just requires writing the test (as a bash file) and adding it to the .travis.yml I can do it.

oscarkraemer:
So it is some time since I wrote this; I think the reason why I did not create a test for fails is because I have never manage to make it fail, so I don't know how it fails. It should be simple to add an additional test for checking for fails. I won't have time to do anything about this before my holidays.

Nortamo:
If its just writing the test and adding it to the  .travis.yml I can fix that, or are there any additional steps required?

oscarkraemer:
It *should* be that simple.

PR #24, opened 2019-05-15
Merge branch 'develop' of github.com:CSCfi/csc-user-guide into make_p…

attesillanpaa:
…outa

separated cloud into pouta and rahti. two versions of TOC available for
your viewing pleasure and comments


attesillanpaa:
Hopefully correct...

PR #39, opened 2019-08-08
Brings all structural changes from develop to master

Nortamo:
**Reason**
The structure and functionality of the master and develop branch
need to match when we start adding material to the master (publicly visible page)
branch from develop.

**Details**
After merging the pull request, the develop and master branch only differ in the `docs` directory
(`img` and `css` are the same) and the mkdocks.yml (Nav bar is empty). The dummy index.md files in the master branch docs subdirectories should be removed when the first content is added to the respective category.


Nortamo:
The Pull request test passes.
The Branch test fails with: 
`Timeout waiting for network availability.` So a problem with travis itself?

This should be ok to merge, but would be good if another pair of eyes would have a look

Nortamo:
The force-push was just to force travis to run the test again, Now all test pass.
Only change in force push was adding one line to TODO.md

PR #58, opened 2019-08-27
switch apps list menu order

attesillanpaa:


mlehtiva:
Seems legit.

nruneberg:
fine

mlehtiva:
Ok still seems legit.

PR #60, opened 2019-08-29
Fix billing

galfthan:
Some fixes to billing page.

Fixed lists
Changed headers to be per service
Fixed project => Scratch disk


Nortamo:
Seems ok.

PR #65, opened 2019-08-30
Example job scripts for GPUs

mlouhivu:


Nortamo:
Looks good. Two questions:

1. Do we want to have the `-l` flag in our examples (i just checked that the flag is currently in use in all the examples.) 
2.  Should the multiple GPUs example have multiple tasks (as it is now) or just one task?

PR #72, opened 2019-09-02
Update system.md

UFOPOLI:


galfthan:
In english the standard is to use  dot (.), and not comma.

UFOPOLI:
The decimal separator (and other similar standards) are dependent on the **domain** (i.e. country), not language. Finland is a comma country, not decimal point.

https://en.wikipedia.org/wiki/Decimal_separator#Arabic_numerals

Nortamo:
Can this be merged?

PR #90, opened 2019-09-03
Instructions for $TMPDIR on login nodes and general about login nodes

Nortamo:
**Do not merge yet!**
I will look at this again tomorrow.

### DIFF
--- a/docs/computing/disk.md
+++ b/docs/computing/disk.md
@@ -137,9 +137,14 @@ is located under `$TMPDIR` and is separate for each login node.
 
 The local storage is good for compiling applications and performing 
 pre- and postprocessing that require heavy IO operations, for example packing and unpacking 
-archive files. Remember to move your data to a shared disk area afterwards. 
+archive files. 
+
+!!! Note
+    The local storage is meant for **temporary** storage and is cleaned frequently.
+    Remember to move your data to a shared disk area after completing your task. 
 
 ### Compute nodes 
 
-The IO- and gpu-nodes have local fast storage available for jobs.
-For more information see: [creating job scripts](running/creating-job-scripts.md#local-storage) 
+The IO- and gpu-nodes have local fast storage available upon request for jobs.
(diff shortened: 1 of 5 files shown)


docs/computing/disk.md
```diff
@@ -126,3 +126,20 @@ quota regardless of how much data you actually have in the scratch
...
+## Additional disk areas
+
+### Login nodes
+
+All of the login nodes have 2900 GiB of fast local storage. The storage
+is located under `$TMPDIR` and is separate for each login node.  
+
+The local storage is good for compiling applications and performing 
+pre- and postprocessing that require heavy IO operations, for example packing and unpacking 
+archive files. Remember to move your data to a shared disk area afterwards.
```
galfthan: is there any cleaning of this data?

docs/computing/system.md
```diff
@@ -40,12 +40,19 @@ A detailed list of the nodes:
...
+    The login nodes can be used for **light** pre- and postprocessing, compiling
+    applications and moving data. All other tasks are to be done using the 
+    [batch job system](running/getting-started.md). Programs not adhering to these
+    rules will be terminated without warning.
 
 
 ### Storage
 
-Puhti has a 5+ PB Lustre parallel storage system providing space for home, project and scratch storages. 
+Puhti has a 5+ PB Lustre parallel storage system providing space for [home](disk.md#home-directory),
```
galfthan: We could fix at the same time. It is actually 4.8 PB...

galfthan:
Looks pretty good already now.

docs/computing/disk.md
Nortamo: Clean up if files have not been accessed in 10 days
`/usr/sbin/tmpwatch "$flags" -x /local_scratch/build 10d /local_scratch`
Should we mention it explicitly or just say that files are frequently cleaned?

docs/computing/system.md
Nortamo: I will change to the correct number

PR #94, opened 2019-09-04
Puhti connect

kkmattil:
One page about connecting Puhti.


Nortamo:
Now Connecting to Puhti is used as a headder twice on the same page  (overview)

Info and instructions about login nodes is already included in #90
So #94 (this) adds Putty and X instructions, not quite sure where is the natural place for these.

Nortamo:
Let's have a look at this after #90 is merged

PR #97, opened 2019-09-04
Initiliazied Elmer documentation

tzwinger:


Nortamo:
The force pushes are just me amending my previous commits

PR #105, opened 2019-09-10
Update index.md

ktiits:
Add geoconda


Nortamo:
The travis test fail as the geoconda.md file is not present i this branch.
I merged #104 which contained the page so this should work.
(Travis does not rerun tests even though the target branch in updated.)

Nortamo:
All tests pass

PR #118, opened 2019-09-19
Style tests

Nortamo:
I added some extra test to check for common mistakes in the slurm batch job scripts.
Test:
-  `check_commands.sh` compares all `#SBATCH` options with the ones available through `sbatch -h` and fails if the documentation contains something which is not present in the `sbatch` output

- `check_partitions.sh` fails if invalid partition names are given. Valid partitions are taken from 
`scontrol show -a partition  | grep Name | cut -d"=" -f2` (The values are hard coded into the test so if we change the partitions , we need to edit the tests)

-  `check_mail_commented.sh` fails if the `--mail-user` flag is not commented out in all batch job scripts. 

- `check_long_slurm_flags.sh` fails if short slurm flags are used.


attesillanpaa:
Manual testing confirmed everything to run ok.

PR #119, opened 2019-09-19
Second version of Accounts section

penikune:
https://csc-guide-preview.rahtiapp.fi/origin/penikune-patch-2019-09-19/#accounts/


penikune:
@jotaskin told me that she has checked the changes and they look fine.

PR #122, opened 2019-09-20
Md link tests

Nortamo:
Implemented some test for markdown links:
First draft, these could probably be improved on.

The test check for (And fail if some are found) the following things:
- file links which don't point to a valid file (this has some overlap
with mkdocs, but also catches links which do not end in .md)
- section links which don't point to a valid section. (the ./site folder
needs to be present for this test to work, so run mkdocs build before
running this test)
- Hidden files, i.e files which are not referenced to at all. If for
some reason a file is intentionally hidden, add it to the
hidden_whitelist.txt. Only hide files for a very good reason!
There are some applications files there at the moment which need to be
completed and linked to ASAP
- Urls which point to docs.csc.fi, any links to docs.csc.fi should be
done using internal links (otherwise checking for validity is much more
difficult)

There is a file for whitelisting files that are intentionally left hidden 'tests/hidden_whitelist.txt'.
(hidden files are files which have no links to them, so they can only be accessed by
typing in the exact url.). **Files should be hidden only with a very good reason.**


Nortamo:
The test should work ok now.  They will probably break if you name your file/sections something really weird using a lot of special characters, in that case inform me and I will have a look.

Nortamo:
A squash and merge is probably the best

Nortamo:
I cleaned the pull request, so it should now be ok to do a normal merge

Nortamo:
There is an issue with "absolute paths in md links"

PR #143, opened 2019-10-03
Added documentation about Julia.

sjsillan:


Nortamo:
The application needs to be added to the software listing.

docs/apps/julia.md
```diff
@@ -0,0 +1,103 @@
...
+function library.
+
+For a quick introduction and tutorial, see https://github.com/csc-training/julia-introduction 
+
+[TOC]
+
+## Available
+
+### Puhti
+-   1.1.1 compiled with Intel Math Kernel Library (MKL)
```
Nortamo: The file appl/soft/math/julia/gcc-mkl/julia-1.1.1/usr has the wrong permissions, and is not accessible

docs/apps/julia.md
```diff
@@ -0,0 +1,103 @@
...
+-   1.1.1 from pre-compiled binaries
```
Nortamo: Some more information for the users could be needed, i.e when should you use which version?

sjsillan: I added recommendation to use pre-compiled binaries, when using ARPACK package.

PR #144, opened 2019-10-03
Allas tutorial

kkmattil:
1st Allas tutorial


galfthan:
This should be also added to the tutorials section (https://csc-guide-preview.rahtiapp.fi/origin/allas_tutorial/#support/tutorials/)  AND also reference from the tutorial on moving to puhti: https://csc-guide-preview.rahtiapp.fi/origin/allas_tutorial/#support/tutorials/puhti_quick/

Did not read through everything else yet.

Nortamo:
The test fail due to manual html anchors, this is ok and we are aware of this.

PR #147, opened 2019-10-03
Reworded Allas client documentation

Aegil:
The flavor word can cause confusion as openstack uses flavors in
reference to instance types. This can cause confusion as flavors in the
Openstack sense don't apply to allas so client / tool is a more
appropriate naming.

Also fixed a typo in the image naming.


docs/data/Allas/introduction.md
```diff
@@ -28,22 +28,22 @@ The stored objects can be of any data type, such as images or compressed data fi
  * Files larger than 5 GB must be divided into smaller segments (normally this is done automatically during upload), see [Files larger than 5 GB](./using_allas/common_use_cases.md#files-larger-than-5-gb).
 
 
-## Different flavors for using Allas
```
galfthan: Or ways? clients is already quite specific

Aegil: reworded.

PR #148, opened 2019-10-03
Add Allas to Quick links on the main page

penikune:
https://csc-guide-preview.rahtiapp.fi/origin/penikune-patch-1/


galfthan:
Looks better than current production one

PR #161, opened 2019-10-07
Update accessing_allas.md

UFOPOLI:


Nortamo:
Some broken links, I will fix them. 
When correcting section names, all the links to them also have to be corrected. 
The tests should report which links have been broken.

PR #165, opened 2019-10-08
Update common_use_cases.md

UFOPOLI:
There have been at least four different ways "a_commands" has been written. I was told that "a-commands" is the right form, is this correct? Somebody who knows the answer needs to go through all of these documents and unify the typography.


kkmattil:
a-commands is better  than a_commands.

PR #205, opened 2019-11-05
Adding pouta documentation

oscarkraemer:
This is should more or less be the same documentation that is already in develop branch. I moved the Pouta FAQ to cloud/pouta instead of support.


attesillanpaa:
Is it a good idea to split the FAQs around the site? I originally moved the pouta FAQ to the common location. There's a separate branch for the new common FAQ but it's missing content. This would be a good start. Opinions?

penikune:
How frequently are the questions in Pouta FAQ asked? If not often, maybe it could be called something else: Tips and tricks, Useful techniques, or something.

oscarkraemer:
If the FAQ is no good, please tell me what to change it to. I have no strong opinions I just want to get rid of Liferay so that we (pouta-team) does not need to be in the limbo we have been in the last 6 month; where we can't update the documentation because we would need to change it in two places.

oscarkraemer:
Also I'm not able to replicate the Travis section links on my computer. Can somebody give me some pointers how to fix it or is Travis wrong?

attesillanpaa:
I'd propose to have the FAQ as it was in the develop branch (index page with TOC, not as individual entries on the left menu). We're about to publish a bunch of other FAQ entries too, and the Pouta FAQ would help a lot with the content. So the TOC categorization is relevant.
> If the FAQ is no good, please tell me what to change it to. I have no strong opinions I just want to get rid of Liferay so that we (pouta-team) does not need to be in the limbo we have been in the last 6 month; where we can't update the documentation because we would need to change it in two places.

oscarkraemer:
I updated the mkdocs.yml so that the "Pouta FAQ" is a part of the main topic FAQ.

Note that the change I did, did not change the directory structure nor the URL of the documents.

attesillanpaa:
I now moved them so that they don't dominate the left menu but form their own section in the main FAQ, which I also moved to under the support main entry. If you want, a link to it might make sense from the pouta contents.md ? We'll soon add some other FAQ entries too and then this FAQ is good to go on my behalf. Did you check the content that it's valid?

oscarkraemer:
Thank you for moving the FAQ.

I went through the content in the develop branch and tried to remove/update anything related to pouta-object-storage. I think that the documentation is at the moment more valid than Liferay.

Regarding adding a link to the content.md, I think it can be added later (by anybody) if needed.

PR #241, opened 2019-11-20
Use client9/misspell to spellcheck

martbhell:
misspell has a nice "-w" argument that was used to fix the spelling errors in
all the changes to *.md files in this commit.

It seems the aspell that was in there has been disabled. Maybe misspell is easier to use? At least to me it seems to have caught a few easily fixed words.

I am also fine with removing running misspell as part of CI and just fixing these typos as it seems that https://github.com/client9/misspell has not been updated since 2018


Nortamo:
The reason the spellcheck was disabled was due to the amount of false positives it was generating.
Anything from puhti,slurm to metadata would trigger an error and you would need to maintain a very large list of words to whitelist.

Does misspell avoid these kinds of issues?

The `-w` options seems ok, I'm just wondering how aggressive it is and what the possibility of breaking something is.

If these things are deemed ok, I am all for automated spelling tests. 

I will do a quick test how the `-w` option affects the whole documentation

martbhell:
> The reason the spellcheck was disabled was due to the amount of false positives it was generating.
> Anything from puhti,slurm to metadata would trigger an error and you would need to maintain a very large list of words to whitelist.
> 
> Does misspell avoid these kinds of issues?

It seems it avoids most of our false negatives. It only cares about a (subjectively) compiled list of commonly misspelled words.

> 
> The `-w` options seems ok, I'm just wondering how aggressive it is and what the possibility of breaking something is.
I found it useful when running the command manually.
> 
> If these things are deemed ok, I am all for automated spelling tests.
> 
> I will do a quick test how the `-w` option affects the whole documentation

This PR does not make travis automatically fix typos in the git. It would just fail.
I'm not sure how to go about having it automatically fix.
That might be automatically doable with a kind of github bot, that every now and then checks for errors and if so submits a PR. I have no experience with github bots/apps so can't say.

Nortamo:
I think we will manage without the `-w` but otherwise using this is a good idea.
I think the test is now crawling through the site folder (not such a big problem) also,

so we could do something like
`misspell --error docs/* *.md *.yml`
(or equivalent)
to avoid that and explicitly state that we are only looking at markdown files. 

Otherwise this looks good and I think we could merge this.
If we run into problems in the future with the misspell we can switch program
or just disable the spelling tests

martbhell:
> so we could do something like
> `misspell --error docs/* *.md *.yml`
> (or equivalent)
> to avoid that and explicitly state that we are only looking at markdown/.yml.

misspell seems a bit funny how to write the path to what should be checked.

I changed to <pre>misspell -error (!mkdocs*)*</pre>

A <pre>find</pre> or some such would be needed to make it only run against .md and .yml files. I could change to that if you want?

Nortamo:
If you have the time to do that it would be nice, otherwise I don't think that it is such a big issue
and we can do the merge with these changes.

martbhell:
np. Updated with a find|xargs combo.

Nortamo:
Great. The pull request looks good, and all tests are passing so I will merge this into the master branch

PR #243, opened 2019-11-20
Clean unused tests

Nortamo:
No functionality is changed, just removed files which were not being used anymore.


Nortamo:
The Branch test fail due to missing grid_blast.md link, this is fixed in the master branch.

PR #247, opened 2019-11-21
Master octave

sjsillan:


attesillanpaa:
perhaps add a tip to load the gnuplot module to enable plotting from octave

PR #248, opened 2019-11-21
X11 interactive example

attesillanpaa:
added an example srun command to interactively run x11-stuff on a compute node.


galfthan:
If email does not work then we should remove it.

attesillanpaa:
It does. There was some minor hiccup which has been resolved. There's a number of other examples also using it in docs, too.

PR #257, opened 2019-11-26
add site_url to build sitemap.xml correctly

juhanikataja:
Discussion on mkdocs issue: https://github.com/mkdocs/mkdocs/issues/1783


juhanikataja:
It appears to generate sitemap now correctly: https://csc-guide-preview.rahtiapp.fi/origin/add-site_url/sitemap.xml

PR #268, opened 2019-11-28
More instructions about interactive usage of Puhti

jussienko:


nruneberg:
Should it also be mentioned here that the interactive graphics command can also be used through NoMachine for faster graphics?

jussienko:
> Should it also be mentioned here that the interactive graphics command can also be used through NoMachine for faster graphics?

That's mentioned in the beginning part of the article, but as there is no x11 in bash example, maybe it would be better to move that to "... with X11' part?

PR #271, opened 2019-11-28
spatial r environment

johannesnyman:


ktiits:
Add references from Taito page. Latest link to our course is: https://www.csc.fi/web/training/-/spatial-data-analysis-with-1
Move fasterize to alphabetic order.

PR #273, opened 2019-11-29
 Fixes exit code of spellcheck test

Nortamo:
Travis tests were not failing even though the test reported misspelled words.


Nortamo:
One review is sufficient for merge.

martbhell:
I have strong recollection of using the "|xargs" way because it worked better than the "find - exec" in  regards to return codes.. I wonder if there's still some corner case. Oh well, I guess time will tell :)

PR #279, opened 2019-12-04
Update sla.md

UFOPOLI:


attesillanpaa:
the automatic build fails due to anchor in an internal link. However, it works, so I'll force accept it once the content has been reviewed.

PR #285, opened 2019-12-12
Add FAQs from research.csc.fi

mlehtiva:
Add possibly relevant FAQs from https://research.csc.fi/faq-details.


mlehtiva:
Fixed broken links -ready for production.

PR #302, opened 2019-12-18
Draft: Sugestion how to utilise markdown-include

oscarkraemer:
How it works:
https://github.com/cmacmackin/markdown-include

Selling feature: 
`provides an "include" function, similar to that found in LaTeX `

Possible problem: 
* It works locally after `pip install markdown-include` but it does not seem to work in the community csc-guide-preview.rahtiapp.fi . IDK if the markdown-include needs to be installed by an admin.
* There is one test that fails because of "hidden files"


oscarkraemer:
If you think this is a horrible idea please say so and close the Draft PR.

jlento:
In general, I would keep the markdown extensions to absolute minimum for maximum portability of the content "source code". Macros in the static site generators' templates, little bit of javascript, and plain old hypertext links could be used instead of includes in the markdown source? Do you have a specific case in mind, which would benefit from markdown source includes?

oscarkraemer:
Hi @jlento 
The Draft "includes" (pun intended) an example. Basically I would like to have the "Pouta" section of support/faq in cloud/pouta without having the rest of the non-pouta FAQ in /cloud/pouta. I would like to be able to do this for Tutorials in the future as well. This would make it possible to have "snippets" in multiple places. It could also be a nice to maintain big tables (e.g. pouta-flavors) in their own files.

There are some drawbacks with the includes.
- Links can't be relative.
- Section depth is not inherent (ATM)

oscarkraemer:
I'll close this since there were no interest.

PR #332, opened 2020-01-17
Change into Material theme

attesillanpaa:
This is a major change on outlook. Links with # in them won't work afterwards, but google indexing will.


Nortamo:
- [X] All test pass
- [X] Embedded videos behave like normal.
- [X] Images, tables, code blocks seem to display properly 

- Also works locally using the `requirements.txt` (centos)

### Some things to note, but not worth changing now. 
- The edit button is convenient for us, but it could possibly be moved to a bit less visible place.

- Not sure why, this pull request seems to affect the loading behavior of the site (Observed with firefox but not with chromium so it could be some browser specific thing. Both preview and local build) .

Changing the page sometimes (quite frequently) reloads the whole page (including the nav-bar) which manifest as white flash when all the content is loaded in. This behavior is not observed with the  
current `docs.csc.fi`. Customers probably won't change page frequently enough for this to be relevant but still worth noting.

Nortamo:
So it seems that with the old version on both firefox and chromium changing page on `docs.csc.fi` only updates the iframe where the article content is. In the new version the page is reloaded everytime

PR #356, opened 2020-01-25
fix sitemap, add robots.txt

attesillanpaa:


nruneberg:
There seems to be one #-link left  from https://csc-guide-preview.rahtiapp.fi/origin/robots/cloud/pouta/known-problems/ to https://csc-guide-preview.rahtiapp.fi/support/faq#pouta -> 404

nruneberg:
If this is only a problem with the preview and absolute links then it's fine!

PR #357, opened 2020-01-27
C pouta sshfs puhti patch 2

kkmattil:
Ohje päivitetty Taidosta Puhtiin.

### DIFF
--- a/docs/support/faq/how-to-access-home-wrkdir-projdir-from-epouta.md
+++ b/docs/support/faq/how-to-access-home-wrkdir-projdir-from-epouta.md
@@ -1,26 +1,32 @@
-# How to access my CSC home/wrkdir/project directories from ePouta?
+# How to access directories in Puhti and Mahti from cPouta?
 
-It's possible to access your regular CSC $HOME, $WRKDIR and even application directories from cPouta in the same way it is possible to access them from outside CSC. These directories are not directly accessible to cPouta because the HPC systems network is separated for security and performance reasons. However if you install `sshfs` tool to you virtual machine you can create a temporary remote mounts for your personal directories in Taito or Sisu.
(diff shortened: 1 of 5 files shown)


docs/support/faq/how-to-access-home-wrkdir-projdir-from-epouta.md
```diff
@@ -1,35 +1,42 @@
...
 
 ```
-sshfs kkayttaj@taito.csc.fi:/wrk/kkayttaj ./taito_wrk
+sshfs kkayttaj@puhtí.csc.fi:/scratch/project_2012345 ./puhti_scratch
 ```
 
-After this, directory _taito_wrk_ shows the content of $WRKDIR in Taito and it can be used as any mounted directory. However, the I/O performance of the remotely mounted directories is not as good as locally mounted directories. Because of that, in the case of I/O intensive tasks it may be more reasonable to copy the data to the local disks of the virtual machine. Tools like `scp` and `rsync` can be used for that.
+(Here, you should replace /scratch/project_2012345 with the scratch directory of your project).
+
+After this, directory _puhti_scratch_ shows the content of scracth in Puhti and it can be used as any mounted directory. However, the I/O performance of the remotely mounted directories is not as good as locally mounted directories. Because of that, in the case of I/O intensive tasks it may be more reasonable to copy the data to the local disks of the virtual machine. Tools like `scp` and `rsync` can be used for that.
```
oscarkraemer: Here is a typo that travis found: `scracth`

attesillanpaa: Is this kind of usage contributing to the disk issues we've been having? I'm in favor or at least pointing to an Allas-route alternative.

PR #358, opened 2020-01-27
Added comments about Conda on Lustre

jlento:


docs/support/tutorials/conda.md
```diff
@@ -57,6 +55,36 @@ from sources, such as [Spack]. Those are not as widely used as Conda, and often
...
+    pip install --user ...
+    ```
+    
+    for python.
+
+2. Use containers
+    
+    This is a great alternative for developing software locally on a workstation,
+    and then deploying it on other workstation, cluster, or on cloud platforms.
+    Puhti supports singularity containers, which are are just single big
```
Nortamo: Should we link to some instructions or do we just leave this for the user to figure out?

jlento: Links would be nice! Please feel free to add links! Although, let's not postpone publishing this, we can add links a bit later too, if not now. I think Ari-Matti has (?) "how to build singularity containers from scratch" instructions cooking.

PR #396, opened 2020-02-14
Update the public bucket example URL

junousi:
- Change object.pouta to a3s.fi
- Add https prefix
- I don't understand what this statement means so I removed it for clarity:
  "it can be shared only to the wanted audience."


attesillanpaa:
The new text mentions the project identified in UUID format. That's new to me. Would a link to wikipedia or some part of our own documentation help here?

junousi:
> The new text mentions the project identified in UUID format. That's new to me. Would a link to wikipedia or some part of our own documentation help here?

Fair point! Added a link, this is currently documented under s3cmd. Project UUID is an OpenStack Keystone term and in a way would deserve a more generic page. Something for another day perhaps.

PR #415, opened 2020-02-25
Upgrade bleach to version 3.1.1

penikune:
Version was 2.1.3.


penikune:
This is because of the security alert:
"We found a potential security vulnerability in one of your dependencies."

shukapoo:
Verified userguide builds up with new bleach version, approving

PR #433, opened 2020-03-03
Move CSC Linux guide from research.csc.fi to docs

mlehtiva:
Chapter 2 "Linux basics for CSC" guide copied from https://research.csc.fi/csc-guide-linux-basics-for-csc  and cleaned.


Nortamo:
The test fail as they do not account for the possibility of an image caption. A fix for this will be pushed to the master branch shortly.
One of the links is a false positive ans is inside ** ** , seems to be a weird grep example
`**[grep](..support/tutorials/env-guide/commands-for-data-processing#grep-selects-rows-that-match-a-given-string)**`

This should be fixed to something witch is not formatted as a markdown link

PR #435, opened 2020-03-03
Added page for TopHat

amsaren:


Nortamo:
There is an invalid slurm option:
`docs/apps/tophat.md:28:#SBATCH --jobname tophat`
`--job-name` is the correct one

PR #436, opened 2020-03-03
Fix image caption link test

Nortamo:
Working link tests when using image captions
    
    Image captions where being parsed as part of the
    link name, which broke the links. Now the
    caption is removed before checking the link


Nortamo:
This was not from the master branch

PR #437, opened 2020-03-03
Fix links using image captions

Nortamo:


Nortamo:
Addresses failed test in #433

PR #439, opened 2020-03-04
Ssh keys

jussienko:


jussienko:
Maybe Atte could check Windows instructions and Juha Linux ones?

PR #443, opened 2020-03-06
Deprecate tb.4 flavor

junousi:
Also remove a slightly confusing sentence about removing flavors
"at some point" in the future.


Aegil:
Changes look good to me from cloud perspective

PR #463, opened 2020-03-24
Kkmattil blast+xhmm 3

kkmattil:
xhmm added + blast updated

### DIFF
--- a/docs/apps/blast.md
+++ b/docs/apps/blast.md
@@ -68,7 +68,7 @@ Would run the same search as described above, except that the  e-value threshold
 
 ## Usage of pb (Parallel BLAST) at CSC
 
-If your query sequence set contains less than 20 sequences then interactive baych job is probably the most effective way top do the search. However, if your query set contains hundreds or thousands of sequences then utilizing the parallel computing capacity of Puhti is more effective. For this kind of massive blast searches you can utilize the `pb` command.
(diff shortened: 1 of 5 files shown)


docs/apps/blast.md
```diff
@@ -35,50 +35,59 @@ Other blast commands
...
 blastp -query proteinseq.fasta -evalue 0.001 -db uniprot -outfmt 7 -out result.table
 ```
-Would run the same search as described above, except that the  e-value threshold would be set to 0.001(-evalue 0.001) and the out put is printed out a a table (-outfmt 7).
+Would run the same search as described above, except that the  e-value threshold would be set to 0.001(-evalue 0.001) and the out put is printed out to a table (-outfmt 7).
+
 
+## Usage of pb (Parallel BLAST) at CSC
 
-## Usage of pb (Parallel BLAST)  at CSC
+If your query sequence set contains less than 20 sequences then interactive baych job is probably the most effective way top do the search. However, if your query set contains hundreds or thousands of sequences then utilizing the parallel computing capacity of Puhti is more effective. For this kind of massive blast searches you can utilize the `pb` command.
```
amsaren:
```suggestion
If your query sequence set contains less than 20 sequences then interactive batch job is probably the most effective way top do the search. However, if your query set contains hundreds or thousands of sequences then utilizing the parallel computing capacity of Puhti is more effective. For this kind of massive blast searches you can utilize the `pb` command.
```

PR #466, opened 2020-03-27
Expand the oc client installation section. 

lvarin:
Even though most of the information is already there, the word "install" was not mentioned, and it relied on the costumer having already an account in Rahti. This way we can point costumers (present and future) to a specific section of the documentation with a clear list of steps.


shukapoo:
LGTM

PR #469, opened 2020-03-30
Training material

mlehtiva:
Add a training material page under the Tutorials section.
Here we can collect the training material links (domain specific material collections elsewhere, MOOCs, webinar recordings, course slides&exercises, Notebooks etc).


attesillanpaa:
branched from ancient state

PR #484, opened 2020-04-07
Bio-tutorials

yetulaxman:
Bioapplications as singularity containers


yetulaxman:
> With this instruction I was able to run singularity in Puhti! (First time for me)
> So this works. Thanks.
> 
> I would like add to the instrctiosn that cPouta can be used to host a machine to do docker->singularity conversions:
> Suggestion. Replace:
> " Please note that one has to do these image conversions in your local machine as Puhti does not grant root access to users."
> Replace
> " Please note that one has to do these image conversions in your local machine or in a virtual machine in cPouta, as Puhti does not grant root access to users."
> 
> Then another suggestion. For me this concept was easier to grasp, when I tested it in an interactive batch job. Maybe you could add that example too:
> 
>     1. Download data:
>        wget https://a3s.fi/pilot_projects/Deepvariant_singulairty.zip
>        unzip Deepvariant_singulairty.zip
> 
>     2. Start interactive batch job
>        sinteractive -i
> 
>     3. Start interactive singularity session:
>        cd  Deepvariant_singulairty
>        export TMPDIR=$PWD
>        singularity shell  --nv -B $PWD:/data deepvariant_8_0_cpu.simg
> 
>     4. Execute the actual workflow commands in the singularity session
> 
> 
> export PATH=/opt/deepvariant/bin:$PATH
> cd /data/testdata/
> run_deepvariant --model_type=WGS --ref=ucsc.hg19.chr20.unittest.fasta --reads=NA12878_S1.chr20.10_10p1mb.bam --regions "chr20:10,000,000-10,010,000" --output_vcf=output.vcf.gz --output_gvcf=output.g.vcf.gz
> 
>     1. Exit from the singularity session and interactuive batch  job
> 
> 
> exit
> exit

Thanks, Kimmo. Nice suggestion. Have not thought about running deepvariant as interactive job. we can include this part as well.

yetulaxman:
The renaming of zip file (and its folder) was done

PR #508, opened 2020-04-27
Datasets

attesillanpaa:
add a list of datasets on puhti


mlehtiva:
This is good starting point, the common nominator being "datasets + Puhti". To be added: Chipster genome indexes.
(Larger list of all kinds of mixed databases and tools is here: https://research.csc.fi/rdm-services -this list could use some clarification/order.)

PR #510, opened 2020-04-27
Kkmattil chipster genomes 1

kkmattil:
chipster_genomes documentation.


mlehtiva:
Yes! This can then be linked here: https://docs.csc.fi/data/datasets/

mlehtiva:
Merged with master and added the chipster-genomes to https://docs.csc.fi/data/datasets/

PR #582, opened 2020-06-29
WIP: Patch 3

csccva:
Small section about profiling and debugging.


csccva:
Work in Progrerr

mlouhivu:
Looks good, but before merging with the master, I think these should be first merged with the two existing branches: https://github.com/CSCfi/csc-user-guide/tree/debugging and https://github.com/CSCfi/csc-user-guide/tree/profiling.

I can do the merge, so we have a single WIP branch that we can iterate over.

PR #587, opened 2020-07-07
Update GPAW example scripts

jussienko:


Nortamo:
Only `FIXME` test is failing, this is ok in the mahti-dev branch. Merging

PR #589, opened 2020-07-10
Expand documetantion about storage in Rahti

lvarin:
It expands a bit the current storage solution for Rahti. It talks about emptydir, pvcs and their types. Also it mentions Allas.

### DIFF
--- a/docs/cloud/rahti/tutorials/patterns.md
+++ b/docs/cloud/rahti/tutorials/patterns.md
@@ -55,7 +55,7 @@ spec:
 
 EmptyDir provides ephemeral temporal storage locally on the node where containers are run, on Rahti this is RAID-0 SSD storage. When the Pod is killed, the information in the emptyDir will be lost. It is declared directly in the Pod definition:
 
-*`podWithEmptydDir.yamp`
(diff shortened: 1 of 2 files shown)


khappone:
Looks good. In the longer run we should clear up the doucmentation a bit and have e.g. separate chapters for storage, networking, pods, etc. But for now it's better than it was, and has more info on storage which was the point.

Ok for me.

docs/cloud/rahti/tutorials/patterns.md
```diff
@@ -51,6 +51,36 @@ spec:
       claimName: testing-pvc
 ```
 
+## EmptydDir
+
+EmptyDir provides ephemeral temporal storage locally on the node where containers are run, on Rahti this is RAID-0 SSD storage. When the Pod is killed, the information in the emptyDir will be lost. It is declared directly in the Pod definition:
+
+*`podWithEmptydDir.yamp`
```
shukapoo, line 58: .yaml

lvarin: It is fixed, thanks

PR #600, opened 2020-07-28
Debug profile

csccva:


csccva:
Thanks for fixing the docs. They look pretty good now. 
Regarding the tmux Phoung has some notes written in the gitlab. 

Cristian

PR #604, opened 2020-07-31
Add instructions for converting images

matteopozza:
Taking snapshot of virtual machine instances is a desired feature, but
the images obtained are in raw format, which quickly deplete the space
available for images. Alternative and more efficient formats do exist,
but customers do not have instructions on how to convert snapshots using
these formats, and thus saving space ultimately.

This commit will:
- add a section in the page about operations on images. This sections
describes a procedure for converting a snapshot from raw format to qcow2
and upload it to OpenStack.

### DIFF
--- a/docs/cloud/pouta/adding-images.md
+++ b/docs/cloud/pouta/adding-images.md
@@ -1,4 +1,4 @@
-# Creating, uploading and sharing virtual machine images
+# Creating, converting, uploading and sharing virtual machine images
 
 This article explains how to manage images in Pouta.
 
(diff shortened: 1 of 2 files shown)


docs/cloud/pouta/adding-images.md
```diff
@@ -173,6 +173,128 @@ you need to add the line "*acpiphp*" to */etc/modules*. For other
 distros, please check how to load *acpiphp* on boot from the distro
 documentation.
 
+
+## Converting images
+
+When performing a snapshot of a virtual machine, OpenStack creates an 
+image in _raw_ format. These images typically take as many GB as the
```
junousi, line 180: This might not always be the default behavior, in our case it's something we have configured explicitly due to the dataloss bug

https://github.com/CSCfi/cccp/blob/master/manifests/role/compute.pp#L78

junousi: I dunno, it might be fine as is, or could throw in a disclaimer like `Pouta OpenStack is configured to create`

junousi: Or just 

```
-OpenStack creates an image in _raw_ format.
+the resulting image is created in _raw_ format.
```

Up to you, could also be I'm overthinking a bit :)

docs/cloud/pouta/adding-images.md
```diff
@@ -173,6 +173,128 @@ you need to add the line "*acpiphp*" to */etc/modules*. For other
...
+terminal.
+
+    screen -S converter
+
+We now issue the command to download the image obtained from the 
+snapshot.
+
+    openstack image save --file myVmSnapshotRaw.raw <id_of_myVmSnapshot>
+
+Given the size of the image, the process will take few minutes. We can
```
junousi, line 267: If it's only a few minutes, IMO screen is overkill. Have you encountered some scenario where it takes much more time? Then of course it makes sense.

PR #610, opened 2020-08-07
Kkmattil patch 1

kkmattil:
Ipyrad and spades updates


kkmattil:
Just some version updates

PR #632, opened 2020-08-20
Ubuntu-20.04 image will be included in Pouta

khabiruddin:
We are going to provide Ubuntu-20.04 image for our Pouta customers.

CCCP-3301


docs/cloud/pouta/images.md
```diff
@@ -88,6 +88,12 @@ do not want to use CentOS. Note that Ubuntu-18.04 is that one image that
 has `ubuntu` as the default username instead of `cloud-user`, since it was the
 first image type we started to provide directly from the upstream.
 
+### Ubuntu-20.04 LTS
+We will provide Ubuntu-20.04 LTS images which is the latest in Ubuntu image family for who
```
junousi, line 92: Using future tense sounds weird, I would suggest trimming this to:

```
-We will provide Ubuntu-20.04 LTS images which is the latest in Ubuntu image family for who
+Ubuntu-20.04 LTS image is the latest in Ubuntu image family for those who
```

(also `those` added)

junousi:
After this change, the following statement would be untrue:

`Note that Ubuntu-18.04 is that one image that has ubuntu as the default username`

junousi (changes requested):
Couple of fixes plz

khabiruddin:
> After this change, the following statement would be untrue:
> 
> `Note that Ubuntu-18.04 is that one image that has ubuntu as the default username`
> 
> So it needs changing too.

This is not my edition but I would like to change it.

docs/cloud/pouta/images.md
khabiruddin: There are some grammatical fixes.

khabiruddin:
> Couple of fixes plz

Yes I have fixed those.

khabiruddin:
> After this change, the following statement would be untrue:
> 
> `Note that Ubuntu-18.04 is that one image that has ubuntu as the default username`
> 
> So it needs changing too.

Yes done.

junousi:
One more thing - can you amend Ubuntu 20.04 to the table on the same page, too

khabiruddin:
> One more thing - can you amend Ubuntu 20.04 to the table on the same page, too

Yes done.

junousi (approved):
Looks good to me!

PR #640, opened 2020-08-24
Added hyperthreading

samiilvonen:


jussienko (approved):
Does it make sense to specify only --nodes for pure MPI jobs and rely on default --ntasks-per-node=128 ?

PR #642, opened 2020-08-24
Added new suggested compiler flags

samiilvonen:


jussienko:
With gcc 9, -march=native is equivalent to -march=znver2 and triggers also -mtune=znver2

PR #643, opened 2020-08-24
removed reference to pilot period & changed servicedesk to mycsc

heikonenj:


attesillanpaa:
the whole file has been removed, so this PR is obsolete

PR #657, opened 2020-08-26
remove OMP_PLACES from example

attesillanpaa:


jussienko:
7.1-elpa works correctly (and better) with OMP_PLACES

PR #658, opened 2020-08-26
Csccva last minute patch

csccva:
Specified that he tools are available only on Puhti


docs/apps/vtune.md
```diff
@@ -1,7 +1,6 @@
 # Intel VTune Profiler
 
-Intel VTune Profiler is provided via the `intel-vtune` module. One sets up the
-environment by loading the module:
+Intel VTune Profiler is available only on Puhti and it is provided via the `intel-vtune` module. At this moment it support only the applications compiled with Intel MPI library. One sets up the environment by loading the module:
```
Nortamo, line 3: Is the "Intel MPI library" requirement also necessary   if you want to profile something other than the MPI behavior?

Nortamo:
The applications are missing from the software listing in `docs/apps/index.md`

docs/apps/vtune.md
jussienko: At least MPICH works also, about hpcx (OpenMPI) I do not know but it should be at least tested.

Nortamo: Single process profiling works as expected for a test application compiled with `hpcx-mpi`

csccva: Hello,

thank you for looking over the documentations. I misunderstood the term "support" on the Intel documentation page.. That sentence can be removed.

csccva:
I will fix this in the next branch. I am adding more to the profiling. The same for ITAC. 

What does qualify as an application. For example should nvprof or mpiP be put i the list of Applications?

Nortamo:
> I will fix this in the next branch. I am adding more to the profiling. The same for ITAC.

Ok, I'll fix the vtune sentence and merge this pull request. 

> What does qualify as an application. For example should nvprof or mpiP be put i the list of Applications?

I would say that any piece of software which warrants it's own page can be included in the list of applications, This makes it easier for customers to find and discover the applications. If we have several debugging and profiling applications we could create a separate header for those  in `docs/apps/index.md` and just refer to that list from the Debugging and Profiling pages under computing.

I would add nvprof and mpiP to the application listing.

csccva:
I will try to finish the profiling tools missing and after move everything to applications.

Nortamo (approved):

PR #666, opened 2020-08-27
lammps compile instructions advertised

attesillanpaa:
even this is not available as a module I'd say it makes sense to advertise how to compile it


nruneberg (commented):
Please don't use the prebuilt binries -> binaries

nruneberg (approved):
Ok

PR #668, opened 2020-08-28
Search bar dummification

johannesnyman:
Removed "-" separator from search and disabled some supposedly smart search features. Took way too much time as the documentation guided to change mkdocs.yml which did absolutely nothing. Key was messing with the localization file that had quite a lot of hidden config parameters. Found them after finding this comment

https://github.com/squidfunk/mkdocs-material/issues/1099#issuecomment-491503493

I think it works better now. 

a-commands work
#SBATCH returns batch scripts


attesillanpaa (approved):
Brilliant!

PR #672, opened 2020-08-31
Add documentation on how to send emails in Rahti. 

lvarin:
This is just to add documentation on how to send emails in Rahti.

This documentation is taken from Pouta's docs. It is copied over to Rahti's to give visibility to Rahti users.


tfrisk (approved):

lvarin:
I will add an example, so the headers to use are more clear.

lvarin:
Ready to review

joasilva:
Looks good python-wise and can actually be quite useful of our users :-)

joasilva (approved):
Looks good python-wise and can actually be quite useful of our users :-)

PR #673, opened 2020-08-31
Csccva more profiling

csccva:
Adding new profiling for Mahti. mpiP


csccva:
mpip for Mahti and Puhti

docs/computing/mpip.md
```diff
@@ -0,0 +1,198 @@
...
+```
+The path to the iberty library is `/appl/spack/install-tree/intel-19.0.4/libiberty-2.31.1-o4es74/lib/`, while the path to unwind library is `/appl/spack/install-tree/intel-19.0.4/libunwind-1.2.1-45uplb/lib/`. 
+The above re-link will only work if it appears the last in the compiling line. 
+Next the code is ran as a usual batch job. The follwing additions are needed to the job script:
+```
+module load mpip
+export LD_LIBRARY_PATH=<path-to-unwind-lib>:<path-to-iberty-lib>:$LD_LIBRARY_PATH
+```
+For running instead we replace `srun` by `srun -ppdebug`. 
+This will create profiling for all code in a file which is file which indicate in the standard output before the programs's own output. Here is an example of a output at the beginning of the execution of a code:
```
gmarkomanolis, line 24: Probably typo:
"This will create profiling for all code in a file which is file which" should be "This will create profiling for all code in a file which"

gmarkomanolis (commented):
I added two comments, otherwise seems good-

docs/computing/mpip.md
```diff
@@ -0,0 +1,198 @@
...
+-lmpiP -lm -L<path-to-iberty-lib>  -liberty -L<path-to-unwind-lib> -lunwind
+```
+The path to the iberty library is `/appl/spack/install-tree/intel-19.0.4/libiberty-2.31.1-o4es74/lib/`, while the path to unwind library is `/appl/spack/install-tree/intel-19.0.4/libunwind-1.2.1-45uplb/lib/`. 
+The above re-link will only work if it appears the last in the compiling line. 
+Next the code is ran as a usual batch job. The follwing additions are needed to the job script:
+```
+module load mpip
+export LD_LIBRARY_PATH=<path-to-unwind-lib>:<path-to-iberty-lib>:$LD_LIBRARY_PATH
+```
+For running instead we replace `srun` by `srun -ppdebug`.
```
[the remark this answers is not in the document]
Nortamo: `-ppdebug` is not a valid slurm flag,  as pdebug is not a valid partition. So that should probably be removed / replace with --partition=test or similar.

docs/computing/mpip.md
```diff
@@ -0,0 +1,198 @@
+# mpiP: Lightweight MPI Profiler
+
+[mpiP](http://mpip.sourceforge.net/) is a lightweight MPI profiler which uses statistical sampling to record profiling data. At the momenmt is only available on Mahti. It generates less overhead and much less data than tracing tools. In order to use no code changes are require, only a re-link is needed. For both Mahti and Puhti, in addition to the required modules, one has the load the mpiP module:
```
Nortamo, line 3: The module is available both on mahti and puhti?

docs/computing/mpip.md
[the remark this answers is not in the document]
csccva: > I tried without -ppdebug and it works, not sure why to need -ppdebug, is there a reason?

I think I took that line form the manual.

docs/computing/mpip.md
```diff
@@ -0,0 +1,198 @@
...
+Next the code is compiled as normal, but with addition of the `-g` flag and the link to the libraries. For Mahti one should use:
+```
+-lmpiP -lm -lbfd -liberty -L<path-to-unwind-lib>-lunwind
+```
+The path to the unwind library on mahti is `/appl/spack/v014/install-tree/gcc-9.3.0/libunwind-1.3.1-otflii/lib`. 
+Similarly for Puhti the relink is done with :
+```
+-lmpiP -lm -L<path-to-iberty-lib>  -liberty -L<path-to-unwind-lib> -lunwind
+```
+The path to the iberty library is `/appl/spack/install-tree/intel-19.0.4/libiberty-2.31.1-o4es74/lib/`, while the path to unwind library is `/appl/spack/install-tree/intel-19.0.4/libunwind-1.2.1-45uplb/lib/`.
```
Nortamo, line 16: Will this work also work with the gcc stack? or does there need to be separate  mpip installation for that stack.
If it works for both it could be better to create a module which sets the proper paths and library flags, 
explicitly depending on a specific software installation in the documentation should be avoided.

docs/computing/mpip.md
csccva: I solved it. It should be ready for merging.

Nortamo (changes requested):
Looks good, 
The availability should be made clearer, preferably in the same format 
as other applications e.g

```
## Available
    Puhti: version
    Mahti: version
```
If at all possible I would like to have those absolute paths behind a module or similar,
but this can be done at a later time.

docs/computing/mpip.md
csccva: This part has to be solved. I put the intel because that is the recommended one on Puhti, but the gcc should work as well. I can see in that the folders are there in the install tree in the gcc folder. The path to the additional libraries should be loaded automatically depending on which modules are loaded.

docs/computing/mpip.md
csccva: Hello, 

Yes. The module is available on both Puhti and Mahti.  I added the availability.

csccva:
> Looks good,
> The availability should be made clearer, preferably in the same format
> as other applications e.g
> 
> ```
> ## Available
>     Puhti: version
>     Mahti: version
> ```
> 
> If at all possible I would like to have those absolute paths behind a module or similar,
> but this can be done at a later time.

Nortamo (approved):

PR #677, opened 2020-09-01
Listing automation

Nortamo:
# IMPORTANT NOTE !
This pull request modifies de dockerfile used for deploying the documentation. 
Very important to test properly before merge.

custom preview (rahti-preview uses the docker image from the master branch so it's not useful for
testing these changes  )
http://user-guide-677-nortamo.rahtiapp.fi/

## Content 
- automatic generation for alphabetic software listing and a by system software listing 
- Last edited time stamps on the bottom of pages, this is to more easily spot old or possibly deprecated information. 

## Current functionality and design decisions
- By availability page added to the nav-bar under Applications
- All system categories share the same page
- The system page start's with the categories listed with an optional description of the system
- New systems can be added, one only needs to specify the name,keyword to search for and an optional description.
- A application can be present on multiple systems. 
- Blacklists exits for generation of timestamps, systems and alphabetic listing
    - e.g no time stamp on the front page. Application pages can be excluded on a per system basis
    - The blacklist if meant for edge cases which should be rare, otherwise we will change the keyword or enforce stricter formats on the application pages.  

## Questions about software listings 
- Is the formatting ok
- One page or separate pages for each system
- Other systems?

### Question about timestamps 
- Do they bring any added value to us or the end user
- If used where should they be located and how should they be formatted 

The build-config for the deployment should not need to be edited as master is the default value for the 
branch to checkout 
`ARG repo_branch=master`


attesillanpaa:
By systems:
* I like the format
* I'd use one page for now
* Other potential systems would be: notebooks?, Rahti/Pouta/kvasi/...?

Timestamp:
* This was suggested (can't remember who it was, sorry) to indicate the end user that the data is up-to-date
* I like the location and format

jussienko:
If one wants to check the systems only from **## Available** section, the following grep should do the job (note also that with -l awk, sort and uniq should not be needed):
```
files=$(grep --exclude-from=<(echo "$ignores") -Pzil "(?<=^## Available)\s*\n[^#]*${system_key[$i]}" $app_dir/*.md | sed s%docs/apps/%%)
```
If checking only the **Available** section, there should probably be also a check that section exists and contains something sensible...

attesillanpaa:
I would like to merge this. If we want to change the way availability (and I'd like to) is grepped, we can do it later, but first the app entries should edited to conform to the policy.

jussienko:
> I would like to merge this. If we want to change the way availability (and I'd like to) is grepped, we can do it later, but first the app entries should edited to conform to the policy.

Yep, there are quite few apps that require that, e.g.
```
grep -lPi '(puhti|mahti)' docs/apps/*md | xargs grep -PzLi "(?<=^## Available)\s*\n[^#]*(puhti|mahti)"
```

attesillanpaa (approved):
Very welcome update!

PR #684, opened 2020-09-04
Update scalability test instructions

jussienko:


docs/accounts/how-to-access-mahti-large-partition.md
```diff
@@ -31,15 +31,12 @@ not affect results, typically few minutes for the shortest run time
 (largest node count) is fine.
 * Parameters affecting the scalability can, and are encouraged to be,
 changed. Note, also the [performance checklist](../computing/running/performance-checklist.md)
+* Scalability is considered good if the parallel efficiency is 85 %, minimum requirement is 75 % efficiency
```
galfthan, line 34: Do we need to give such an exact number for what is good? Perhaps just state 75% is minimum but one should target higher. 

Also, should we state these in terms of speedup when doubling number of nodes, imho parallel efficiency is more difficult to understand and is not described here.

mlouhivu: Yes, I agree that the 85% should be dropped.

About the second point, speedup is already used as an explanation for the parallel efficiency (in the next line), so IMHO it's all good.

docs/accounts/how-to-access-mahti-large-partition.md
```diff
@@ -31,15 +31,12 @@ not affect results, typically few minutes for the shortest run time
 (largest node count) is fine.
 * Parameters affecting the scalability can, and are encouraged to be,
 changed. Note, also the [performance checklist](../computing/running/performance-checklist.md)
+* Scalability is considered good if the parallel efficiency is 85 %, minimum requirement is 75 % efficiency 
+ (speedup of 1.5 when doubling the number of nodes).
 
-If the parallel performance of a software is self-evident, e.g. based on
-earlier tests carried out on Mahti or other supercomputer, no further
+If the project is continuation of similar earlier project in Mahti, no further
```
mlouhivu, line 37: I feel this should be rephrased somehow; as it is, it sounds like anything that is a continuation is good. Maybe offer that existing data from Mahti (or elsewhere) can be used to justify the request?

jussienko: My intention for using parallel efficiency was that one could test e.g. with 25, 50, 75, 100, ... nodes, and scaling might be ok from 50 to 75, but not from 50-100. Limiting the cases to always doubling the number of nodes is a maybe a bit restrictive, but I agree that parallel efficiency (if we want to use it) should be clearly defined.

For the case of continuation, it should be somehow made clear that the data should reflect the planned input cases, even if it was existing.

mlouhivu (approved):

PR #687, opened 2020-09-10
minor update

attesillanpaa:
changed pmemd.cuda.MPI to pmemd.cuda as we don't recommend running on multiple GPUs. Also highlighted running only GPU aware binaries in gpu partition.


nruneberg (approved):
ok

PR #690, opened 2020-09-14
Mahti billing

galfthan:
Added interactive partition and re-organized scratch to be separately under each system


jussienko (approved):
Ok!

PR #704, opened 2020-09-24
Update s3_client.md

kkmattil:
obtaining s3 keys

### DIFF
--- a/docs/data/Allas/accessing_allas.md
+++ b/docs/data/Allas/accessing_allas.md
@@ -38,7 +38,7 @@ You can start using Allas with one of the following options. Note that the tools
 * **a-tools for basic use:** (Swift, optionally S3) [Quick and safe: a-commands](./using_allas/a_commands.md)
 * **Advanced functions with rclone:** (Swift) [Advanced tool: rclone](./using_allas/rclone.md)
 * **A wide range of functionalities:** (Swift) [Swift client](./using_allas/swift_client.md)
-* **S3 client and persistent Allas connections:** (S3) [S3 client](./using_allas/s3_client.md#s3cmd-with-supercomputers)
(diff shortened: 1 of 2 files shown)


khappone:
Good updates. The  ** Configuring S3 connection in local computer ** basically has the answer. 

It would be nice if we had a clearer set up for API connections for both S3 and SWIFT. E.g. the users in many cases won't use s3cmd at all. 

This may be a bigger change, and out of scope of this PR but what could be nice to have is a  Allas API access page, with e.g.

SWIFT API
Prerequisites
Credentials & endpoints
Testing

S3 API
Prerequisites
Credentials & endpoints
Testing

kkmattil:
API:s will be done only later on.


> Good updates. The ** Configuring S3 connection in local computer ** basically has the answer.
> 
> It would be nice if we had a clearer set up for API connections for both S3 and SWIFT. E.g. the users in many cases won't use s3cmd at all.
> 
> This may be a bigger change, and out of scope of this PR but what could be nice to have is a Allas API access page, with e.g.
> 
> SWIFT API
> Prerequisites
> Credentials & endpoints
> Testing
> 
> S3 API
> Prerequisites
> Credentials & endpoints
> Testing



> Good updates. The ** Configuring S3 connection in local computer ** basically has the answer.
> 
> It would be nice if we had a clearer set up for API connections for both S3 and SWIFT. E.g. the users in many cases won't use s3cmd at all.
> 
> This may be a bigger change, and out of scope of this PR but what could be nice to have is a Allas API access page, with e.g.
> 
> SWIFT API
> Prerequisites
> Credentials & endpoints
> Testing
> 
> S3 API
> Prerequisites
> Credentials & endpoints
> Testing

khappone (approved):

PR #714, opened 2020-10-05
Billing

penikune:
Add projappl, Pouta and Rahti. Rearrange the content.
I tried to embed the BU calculator, but it was too wide for the MkDocs column.


docs/accounts/billing.md
```diff
@@ -1,9 +1,12 @@
 # Billing
 
+See the [Billing unit and price calculator](https://research.csc.fi/billing-and-monitoring#buc) at research.csc.fi.
 
-## Puhti 
+## Accounting formulas
```
galfthan, line 5: Do we really need another header level here? Everything is under this one, seems superfluous

galfthan (changes requested):

docs/accounts/billing.md
penikune: You are right. I'll remove the extra header.
I added it when I tried to embed the billing unit calculator, but forgot to remove it.

galfthan (approved):

PR #716, opened 2020-10-06
Csccva ipm puhti

csccva:
User guide for IPM tool and some typos fixed in the mpip.
IPM is only available on Puhti.


jussienko (commented):
IPM documention mentions also possibility to use it without even relinking, i.e. by setting LD_PRELOAD. Does this work in our systems?

For mpip, would it be possible to have the module to set LIBRARY_PATH and LD_LIBRARY_PATH, so that the bit lengthy settings in the link line and batch job script would not be needed?

csccva:
> IPM documention mentions also possibility to use it without even relinking, i.e. by setting LD_PRELOAD. Does this work in our systems?
> 
> For mpip, would it be possible to have the module to set LIBRARY_PATH and LD_LIBRARY_PATH, so that the bit lengthy settings in the link line and batch job script would not be needed?
Hello,

I will try the possibility without the relinking for the IPM. Give a few hours.

I do not what to say about having the module to set LIBRARY_PATH and LD_LIBRARY_PATH, this part is made by the people who do installations, the "spack group".

csccva:
> IPM documention mentions also possibility to use it without even relinking, i.e. by setting LD_PRELOAD. Does this work in our systems?
> 
> For mpip, would it be possible to have the module to set LIBRARY_PATH and LD_LIBRARY_PATH, so that the bit lengthy settings in the link line and batch job script would not be needed?

I just checked the dynamic profiling works, but  I assume that link to the library is based on the loaded module. I only tested with the path to the intel compiled library. Is there anyway to extract this from the ipm module?

Nortamo (approved):

PR #725, opened 2020-10-16
Update Pouta pages with urgent needs

matteopozza:
The documentation pages of Pouta are missing information about our
storage options, and they have some outdated information. This commit
helps in the following:
- improves the pages describing our storage options
- removes the misunderstanding about the private ssh keys being
copied when using ssh agent forwarding
- updates to python 3 usage
- clarifies, restructures, and fixes typos


junousi (approved):
Excellent.

PR #730, opened 2020-10-21
Update pouta-openstack-rescue-mode.md

miloudbagaa:
This merge request aims to enhance the update pouta-openstack-rescue a bit by considering the following points:

- Give more explanation about the creation of the rescue instance when you have many images with the same name.
- Originally, the document considers only the rescue of CentOS-7 instance. This update adds a comment on how to consider also a CentOS-7 instance.
- Give some hints on how to fix the issue of already existing entries in ~/.ssh/known_hosts. After the creation of a new instance, it will use the same IP address. In this case, the machine of the user can consider this new host as man-in-the-middle.
-  In the created rescue instance, we could observe DNS resolution problem. Then, the user can be not able to fix the issues (install packages) when switch the root to his original disk. To fix this issue, some useful commands have been added.


oscarkraemer:
- I don't think you want a review from me, I'm out of the loop what this PR is about. Pouta have had major version upgrades since I was an admin. 
- One thing that comes to mind is that `rescue` does only work only on nova-backed instances and not on "cinder-volume" backed instances, at least a couple versions ago. 
- It might be a good idea to improve on the PR-message to make it easier for the review to understand what needs to be review and why it needs to be changed.

miloudbagaa:
>     * I don't think you want a review from me, I'm out of the loop what this PR is about. Pouta have had major version upgrades since I was an admin.
> 
>     * One thing that comes to mind is that `rescue` does only work only on nova-backed instances and not on "cinder-volume" backed instances, at least a couple versions ago.
> 
>     * It might be a good idea to improve on the PR-message to make it easier for the review to understand what needs to be review and why it needs to be changed.

Thank you. I do not know how GitHub commands you as potential reviewer. I will improve the PR-message to make it easy to review.

docs/support/faq/pouta-openstack-rescue-mode.md
```diff
@@ -2,8 +2,6 @@
 
 In case you run into a situation where your kernel upgrade fails for some reason, and your CentOS 7 instance won't come up, you might find these instructions useful. Sometimes this might seem like a failed resize but the real reason is the restart that a resize requires.
 
-## Symptoms
-
```
Aegil, line 6: Why are the subheaders removed?

docs/support/faq/pouta-openstack-rescue-mode.md
```diff
@@ -44,27 +42,50 @@ Check your instance Console Log (web UI: Instances -> <your instance> -> Log)
 
 The log says that the instance couldn't boot because it can't find root "Kernel panic - not syncing: VFS: Unable to mount root fs onunknown-block(0,0)". The fix is to use (some) previous, working kernel. Since you can't boot the server, you have to make the fix to the Volume (boot files) by using another instance.
 
-## How to fix the issue
-
```
Aegil, line 48: ?

docs/support/faq/pouta-openstack-rescue-mode.md
```diff
@@ -44,27 +42,50 @@ Check your instance Console Log (web UI: Instances -> <your instance> -> Log)
...
+Also, if your intance is an Ubuntu instead of CentOS, then select Ubuntu image ID instead. 
+
+3. Make sure that the instance is in rescue mode with <pre><code>openstack server show $INSTANCE_UUID</pre></code>
+
+4. ssh into the instances.
+
+In case that you have this message in the terminal:  <pre><code>WARNING: REMOTE HOST IDENTIFICATION HAS CHANGED!</pre></code>
+Fix it by removing the line of your instance IP address from the file ~/.ssh/known_hosts
+
+An alternative way is the execution of the following command:<pre><code>ssh-keygen -f "/root/.ssh/known_hosts" -R "$INSTANCE_IP"</pre></code>
```
Aegil, line 61: That probably won't work as generally people shouldnt really be sshing from root.  ~/.ssh/known_hosts makes more sense

docs/support/faq/pouta-openstack-rescue-mode.md
```diff
@@ -44,27 +42,50 @@ Check your instance Console Log (web UI: Instances -> <your instance> -> Log)
...
+7. Take a backup of grub <pre><code>$ cp /tmp/mnt/boot/grub2/grub.cfg /tmp/mnt/root/grub.cfg.bak-`date +"%F"`</pre></code>
+
+8. Open `/tmp/mnt/boot/grub2/grub.cfg` with your favorite text editor. Remove the first `menuentry` section. *NOTE:* This might not be the correct solution for your specific problem.
+
+9. In case that your instance has issues with some broken packages or drivers, you can switch to your original root file and fix the problems using the following commands:
+<pre><code>
+$ sudo mv /tmp/mnt/etc/resolv.conf{,.bak}
+$ sudo cp /etc/resolv.conf /tmp/mnt/etc/resolv.conf
+$ sudo chroot /tmp/mnt
+$ source /etc/environment
```
Aegil, line 83: Why do you need to source /etc/environment as you would already be the root user of the mounted filesystem.

docs/support/faq/pouta-openstack-rescue-mode.md
miloudbagaa: Ok. I will update it.

docs/support/faq/pouta-openstack-rescue-mode.md
miloudbagaa: During the tests, I have faced issues related with DNS or default gateway. In the rescue mode, the VM will not have a connection. So, if the customer wants to install software, he will have issues. By sourcing the original environment the connection problem should be fixed.

docs/support/faq/pouta-openstack-rescue-mode.md
miloudbagaa: I had some conflict with other changes. I will fix this issues.

docs/support/faq/pouta-openstack-rescue-mode.md
miloudbagaa: The same.

docs/support/faq/pouta-openstack-rescue-mode.md
attesillanpaa: Is this fixed?

Aegil (approved):

PR #735, opened 2020-10-27
Upgrade versions of requirements

lvarin:
Upgraded the versions of packages in `requirements.txt`. 

May need to make this change as well:
```
+++ b/mkdocs_csc/base.html
@@ -88,7 +88,7 @@
         {{ lang.t('skip.link.title') }}
       </a> {% endif %} {% block header %} {% include "partials/header.html" %} {% endblock %}
         <div class="md-container">
-            {% block hero %} {% if page and page.meta and page.meta.hero %} {% include "partials/hero.html" with context %} {% endif %} {% endblock %} {% if feature.tabs %} {% include "partials/tabs.html" %} {% endif %}
+            {% block hero %} {% if page and page.meta and page.meta.hero %} {% include "partials/hero.html" with context %} {% endif %} {% endblock %}
```
mkdocs was having problems to find `feature.tabs`.

### DIFF
--- a/mkdocs_csc/base.html
+++ b/mkdocs_csc/base.html
@@ -88,7 +88,7 @@
         {{ lang.t('skip.link.title') }}
       </a> {% endif %} {% block header %} {% include "partials/header.html" %} {% endblock %}
         <div class="md-container">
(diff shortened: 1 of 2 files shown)


lvarin:
It seems there is some links to fix

lvarin:
Fixed

attesillanpaa (approved):
looks good!

PR #737, opened 2020-10-27
Turbomole update to v7.5, including Mahti

nruneberg:


attesillanpaa (approved):
changed the bold titles to headers so that they work as anchors in links and screen readers

PR #739, opened 2020-10-29
mahti into site metadata

attesillanpaa:


mkdocs.yml
```diff
@@ -1,7 +1,7 @@
 site_name: Docs CSC
 edit_uri: "https://github.com/CSCfi/csc-user-guide/edit/master/docs"
 site_url: "https://docs.csc.fi/"
-site_description: Instructions and user guides for the CSC supercomputer Puhti, Allas storage, Pouta, Rahti, software list, FAQ and tutorials.
+site_description: Instructions and user guides for the CSC supercomputer Mahti, Puhti, Allas storage, Pouta, Rahti, software list, FAQ and tutorials.
```
penikune, line 4:
```suggestion
site_description: Instructions and user guides for the CSC supercomputers, cloud services, storage and software, including FAQ and tutorials
```

penikune (approved):
I suggest that the server names would be removed, because the list is already long and FGCI and Kaivos are missing from there.

mkdocs.yml
attesillanpaa: I don't think that's a good idea. This is metadata and according to google analytics puhti is one of the main keywords leading to our site. Thus, we should tell google (via site metadata) that these pages are about puhti (and mahti). The metadata is not for human reading as far as I understand. The original list was suggested and approved by Haiyan.

penikune:
```suggestion
site_description: Instructions and user guides for the CSC supercomputers Mahti and Puhti, Allas storage, cloud services Pouta and Rahti, and software – including FAQ and tutorials
```

penikune, line 4:
```suggestion
site_description: Instructions, user guides, tutorials and FAQ for CSC supercomputers, cloud and storage: Mahti, Puhti, Allas, Pouta, Rahti, software
```

PR #745, opened 2020-11-05
Deprecation of r-env and rstudio modules

jessepharrison:
The `r-env` (Spack installation of R) and `rstudio` (RStudio Desktop) modules are being phased out.

- Temporarily replacing documentation for these with a message saying the modules have been deprecated, plus a link to `r-env-singularity`.
- Refrences to `r-env` deleted from the R for GIS page.
- Link to `r-env` replaced with `r-env-singularity` in the Data Analysis Guide (under Tutorials).


attesillanpaa (approved):
great! Interesting side note: now that the available puhti/mahti line is missing, R-env (deprecated info) won't be available on the alphabetic or by machine lists. But this is not an issue.

PR #748, opened 2020-11-08
Update fmi.md

jlento:


docs/accounts/fmi.md
```diff
@@ -44,6 +44,13 @@ between __projappl__ and __scratch__:
 
 The FMI customers can use both the regular Puhti login nodes `puhti.csc.fi`, and FMI specific login nodes `puhti-login1.fmi.fi` and `puhti-login2.fmi.fi`.
 
-Puhti FMI works similarly to the regular Puhti system, the main difference being that FMI projects use the `fmi` and `fmitest` partitions instead of the regular partitions (small, large, etc.). Also, the FMI login nodes are networked through FMI's internal network, making network access somewhat different.
+Puhti FMI works similarly to the regular Puhti system, the main difference being that
+
+1. the FMI projects use the `fmi` and `fmitest` partitions instead of the regular partitions (small, large, etc.),
+2. the FMI login nodes are networked through FMI's internal network, making network access somewhat different, and
+3. the FMI compute nodes have simultaneous multithreading (SMT/hyper threading) enabled, thus most tools, such as `htop`,
```
galfthan, line 51: Should SMT be explained better for  less technical users.  Also, the link location can be a bit missleading since it talks about 256 cores, and on puhti exclusive nodes are not needed. Perhaps add a paragraph with some more description and a puhti-fmi example instead of link?

jlento: Good idea! I'll return to this when I have time. Merging now what I have so that there is some documentation.

PR #774, opened 2020-11-25
Replace password with token and add troubleshooting section

lvarin:


joasilva (approved):
Nice that we're advising to use tokens instead of password.

PR #775, opened 2020-11-26
Remove mail edit test

Nortamo:
- `--mail-type` will now work without `--mail-user`. `--mail-user` can be used to override default email. 
- Removed  `--mail-user` from examples and commented out `--mail-type` (Also tests now check for this. )
- Short script to run tests locally, might require installing some small utilities like misspell. 
- FAQ entry about the mail

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,3 +2,4 @@
 *.bak
 *~
 site/
+*.log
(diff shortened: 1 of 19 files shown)


attesillanpaa (approved):
brilliant!

PR #807, opened 2020-12-30
Accessibility v2

johannesnyman:


attesillanpaa (approved):
Nice! Everything else I tested seems to work but also internet exploder gets stuck in the search box with tab (but N shift-tabs get you to the page content backwards...). Can it be hacked, too, or just ignore?

johannesnyman:
> Nice! Everything else I tested seems to work but also internet exploder gets stuck in the search box with tab (but N shift-tabs get you to the page content backwards...). Can it be hacked, too, or just ignore?

I guess I don't really have the possibility to test IE with this mac. That hack should work for IE as well but apparently then doesn't. I have one idea but if it doesn't work, I would skip IE keyboard compatibility then.

johannesnyman:
I think I found the culprit. IE just doesn't handle key presses the same as the others. Now it should work on IE as well (can't test)

attesillanpaa (approved):
works with IE now, too

PR #813, opened 2021-01-07
Alt text to images

attesillanpaa:
Accessibility requires the information contained in images to be available also for the visually impaired. Screen readers can access the alt-text and title labels. This PR adds a crude howto in the STYLEGYUIDE.md and uses the nomachine app-page as an example.
preview: https://csc-guide-preview.rahtiapp.fi/origin/alt-text/apps/nomachine/


mlehtiva (approved):
All good! Nice examples on the NoMachine page.

PR #815, opened 2021-01-11
Add object lifecycle

plaihonen:
Add instructions for setting up an object lifecycle in allas, using s3cmd.


docs/data/Allas/using_allas/s3_client.md
```diff
@@ -345,3 +345,90 @@ would print out an URL that remains valid for 3600 s (1 h). In this case URL, pr
 ```text
 https://fish-bucket.a3s.fi/zebrafish.tgz?AWSAccessKeyId=78e6021a086d52f092b3b2b23bfd7a67&Expires=1599835116&Signature=OLyyCY14s%2F0HxKOOd108mldINyE%3D
 ```
+
+## Setting up an object lifecycle
+
+In order to delete/expire objects automatically, a lifecycle policy can be set-up to the Allas bucket. Objects in the bucket are treated per the lifecycle policy if matching conditions are found. Matching conditions can be set to a prefix and/or tag(s) within the object. Lifecycle policy is especially well suited for the cases where data needs to be removed as a "maintenance" measure after certain intervals.
+
+<font color="red">**Before setting up the lifecycle policy, please check with your department/team that it correctly represents the retention policy for the data in the project.**</font> (Legal or regulatory constrains).
```
junousi, line 353: HTML style `<font>` in markdown - is this a thing? At least github preview doesn't render it.

Alternatively you could consider

```
!!! note
    something_to_emphasize
```

junousi: Ok I now see it as being used in some other places, but it doesn't seem to render correctly anywhere.

I think it's copypasta from the previous docs repository and as such, should be omitted.

junousi (approved):
Approved with a suggestion to improve the nota bene

docs/data/Allas/using_allas/s3_client.md
plaihonen: html code replaced with other emphasis method

junousi: Any reason not to use the other highlight method for consistency? (It's used in quite a few places)

```
$ grep -R ^\!\!\! .|wc
    138     306    6439
```

In any case this is a super minor thing. Approval stands.

plaihonen: Ahh....  I just realized how that renders in the final docs.
In the editor (and its rendering it did not look serious enough)

junousi:
Now I *double* approve this!

PR #816, opened 2021-01-13
Add some commands to the list of module commands

mlehtiva:
Added and modified the list of "commonly used module commands" according to some lecture material


jussienko (commented):
Naming is a bit inconsistent, i.e. there is *modulename*, *name*, *module1*. Maybe this could be unified?

jussienko (approved):
Mycket bra!

jussienko (approved):
Mycket bra!

PR #821, opened 2021-01-21
Fixes socket numbering in image

Nortamo:
The image https://docs.csc.fi/img/mahti_numa.png shows "Socket 0" twice, instead of "Socket 0 and Socket 1".

Changes the second 0 to a 1

Preview at <https://csc-guide-preview.rahtiapp.fi/origin/num_numa_fix/computing/systems-mahti/>


galfthan (approved):
Looks good!

PR #824, opened 2021-01-22
Add short explanation on where to deploy a database

lvarin:
This MR adds a short explanation about where to deploy a Database in the cloud. There are no details about which database to deploy, or how, as it is very well explained already in the databases documentations.


lvarin:
@attesillanpaa what do you think? I tried to keep it simple but useful

attesillanpaa (approved):

PR #826, opened 2021-01-22
Add short explanation about user/password in Rahti

lvarin:
Add short explanation about which username and which password to use in Rahti


tfrisk (approved):
Good clarification.

PR #831, opened 2021-01-27
Kkmattil rclone 2

kkmattil:


docs/data/Allas/using_allas/rclone.md
```diff
@@ -3,6 +3,7 @@
 This chapter contains instructions for using Allas with [Rclone](https://rclone.org/) in the Puhti and Mahti computing environments. _Rclone_ provides a very powerful and versatile way to use Allas and other object storage services. It is able to use both the S3 and Swift protocols (and many others), but in the case of Allas, the Swift protocol is preferred. It is also the default option on the CSC servers. 
 
 > **WARNING:** Rclone should not be used to copy, move or rename objects **inside** Allas. Rclone provides commands for these operations but they don't work correctly for files larger than 5 GB.
+> **WARNING:** If you use rclone and your data upload process gets interrupted, please remove the partially uploaded object before restarting the upload process. Otherwise rclone has sometimes reports a successful data upload even though not all data has been copied to Allas.
```
junousi, line 6: This would be even better if we could link to a rclone bug tracker and outsource the responsibility, but IIUC there isn't one.

In any case I'd like to clarify the latter sentence somehow, maybe something like:

```
-Otherwise rclone has sometimes reports a successful data upload even though not all data has been copied to Allas.
+In this scenario it is possible for rclone to report successful data upload even if all data has not been copied to Allas.
```

attesillanpaa (approved):

PR #843, opened 2021-02-04
Kkmattil allas encryption

kkmattil:
Allas with encryption example.


gitjopa:
In the rclone example, it should be mentioned in the text that the encrypted bucket remote string "allas:2001659-crypt" must start with the reference to allas. In this example case "allas:". Otherwise the encypted data won't go to allas and for instance if just a bucket name would be entered, the data would go in to a new local directory.

gitjopa:
Option to encrypt the metadata file would be good to have.
Also there shoud be a note about possible good and bad effects if the user leaves it unencrypted

gitjopa (approved):

PR #856, opened 2021-02-09
Add explanation for inline Dockerfile

lvarin:
It just give the command to create build in Rahti with a Dockerfile provided in the command itself


docs/cloud/rahti/images/creating.md
```diff
@@ -161,3 +159,13 @@ oc start-build os-sample-python
 ```
 
 Or using [webhooks](/cloud/rahti/tutorials/webhooks/)
+
+## Using the inline Dockerfile method
+
+It is possible to create a new build using a Dockerfile provided in the command line. By doing this, the `Dockerfile` itself will be embedded in the Build object, so there is no need for an external Got repository.
```
tdeneke, line 165:
```suggestion
It is possible to create a new build using a Dockerfile provided in the command line. By doing this, the `Dockerfile` itself will be embedded in the Build object, so there is no need for an external Git repository.
```

tdeneke (changes requested):

attesillanpaa (approved):

tdeneke (approved):

PR #877, opened 2021-02-24
Heads up that GPU performance needs proof

attesillanpaa:
Precaution to GPU partition abuse by CPU capable code which does not speed up enough.


galfthan (approved):
On point

PR #882, opened 2021-03-08
WIP: Started tests on headders and img title and txt

Nortamo:
Adds tests related to accessibility.

### Link descriptions:
- Images need to have both a title and alt-txt `A picture of a whale`
- Links need to have a title  `Wikipedia article about whales`

### Headers
-  Only one lvl 1 allowed per file
- No jumps in depth allowed, e.g. 1->3

#### Unclear
Should link and images be handled in the same way, i.e both tile and alt-txt required ?


rkronberg:
Out of curiosity, what's the status of this PR? Would be nice to have it in production!

PR #892, opened 2021-03-24
Pouta api access

tdeneke:


docs/cloud/pouta/api-access.md
```diff
@@ -0,0 +1,140 @@
...
+as nova (compute), glance (VM images), keystone (authentication), swift 
+(object storage), cinder (block storage), and neutron (networking). 
+
+OpenStack APIs are RESTful and there are multiple ways to interact with 
+them which include using the command-line tools (e.g., `openstack`) or 
+using direct HTTP requests (e.g., with `curl`) or using one of the 
+client libraries (e.g., the `openstacksdk`). 
+
+In this article, we briefly look into using cURL to make direct HTTP requests 
+to the OpenStack APIs  and move to check how to use the `openstacksdk` to automate
```
lvarin, line 14:
```suggestion
to the OpenStack APIs and move to check how to use the `openstacksdk` to automate 
```
(removed extra space)

docs/cloud/pouta/api-access.md
```diff
@@ -0,0 +1,140 @@
...
+to the OpenStack APIs  and move to check how to use the `openstacksdk` to automate 
+some mundane tasks as a demonstration. 
+
+### Pouta access through cURL
+
+Before we can use cURL or any other client to make API requests we need 
+to set some environment variables which hold our credentials to Pouta. This 
+can be done by running a script that you can download from the 
+[Pouta web interface](https://pouta.csc.fi/dashboard/project/api_access/) 
+after logging in. See more details from [here](install-client.md).
```
lvarin, line 23: Links text should be descriptive of the content they link to, ie: Do not use `here` but `install client`

docs/cloud/pouta/api-access.md
```diff
@@ -0,0 +1,140 @@
...
+      }
+    },
+    "scope": {
+      "project": {
+        "domain": {"id": "'$OS_PROJECT_DOMAIN_ID'"},
+        "name": "'$OS_PROJECT_NAME'"
+      }
+    }
+  }
+}' "$OS_AUTH_URL/auth/tokens?nocatalog" | python -m json.tool
```
lvarin, line 54: `| python -m json.tool` must be removed:

```
$ curl -i   -H "Content-Type: application/json"   -d '
{"auth": {
    "identity": {
      "methods": ["password"],
      "password": {
        "user": {
          "name": "'$OS_USERNAME'",
          "domain": {"name": "'$OS_USER_DOMAIN_NAME'"},
          "password": "'$OS_PASSWORD'"
        }
      }
    },
    "scope": {
      "project": {
        "domain": {"id": "'$OS_PROJECT_DOMAIN_ID'"},
        "name": "'$OS_PROJECT_NAME'"
      }
    }
  }
}' "$OS_AUTH_URL/auth/tokens?nocatalog" | python -m json.tool
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100  1087  100   722  100   365   1885    953 --:--:-- --:--:-- --:--:--  1890
No JSON object could be decoded
```

This is due to the fact that `-i` is used to get the headers. And the headers are not a valid JSON object.

docs/cloud/pouta/api-access.md
```diff
@@ -0,0 +1,140 @@
...
+directory, or `/etc/openstack` directory. For quick test purposes, you can add your password in the `clouds.yaml` as: 
+
+```
+clouds:
+  openstack:
+    auth:
+      auth_url: https://pouta.csc.fi:5001/v3
+      username: "username"
+      password: XXXXXXXXXXX
+      project_id: xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx
```
lvarin, line 103: What is project id?

lvarin: Maybe you should say that all values are prefilled in `cloud.yaml` and one only needs to add the password.

docs/cloud/pouta/api-access.md
```diff
@@ -0,0 +1,140 @@
...
+      project_name: "project_xxxxxx"
+      user_domain_name: "Default"
+    regions:
+    - regionOne
+    interface: "public"
+    identity_api_version: 3
+```
```
lvarin, line 110: I am not sure this example is faster than creating the `secure.yaml` file directly.

docs/cloud/pouta/api-access.md
```diff
@@ -0,0 +1,140 @@
...
+clouds:
+  openstack:
+    auth:
+      password: XXXXXXXXXX
+```
+
+Now, you can run the following simple example which lists the available virtual machine flavors:
+
+```
+import openstack
```
lvarin, line 125: Missing shebang, small detail :)

docs/cloud/pouta/api-access.md
```diff
@@ -0,0 +1,140 @@
...
+
+# Initialize and turn on debug logging
+openstack.enable_logging(debug=True)
+
+# Initialize cloud
+conn = openstack.connect(cloud='openstack')
+
+# list VM flavors
```
lvarin, line 133:
```suggestion
# list VM flavors
```

Remove extra spaces

docs/cloud/pouta/api-access.md
tdeneke: added a descriptive label.

docs/cloud/pouta/api-access.md
tdeneke: Good catch. removed `-i` and added `-v -s`. The piping now works fine.

docs/cloud/pouta/api-access.md
tdeneke: Project id is an OpenStack identifier. It comes with the default `clouds.yaml` that can be downloaded from the UI. I think how to add the password is mentioned in the text and it would be clear for the user once they opened the `clouds.yaml` for editing.

docs/cloud/pouta/api-access.md
tdeneke: I think it is good to show both options and let the user make the choice. The `secure.yaml` is also an optional so keeping the text as is makes that fact clear.

docs/cloud/pouta/api-access.md
tdeneke: May be some users want to use the snippet in their functions or test it our in an interpreter so we can leave that detail for the user to choose.

docs/cloud/pouta/api-access.md
tdeneke: Fixed.

docs/cloud/pouta/api-access.md
lvarin: I do not agree. What is the advantage of the first option? That users will be able to commit a file with a password by mistake?

docs/cloud/pouta/api-access.md
lvarin: The interpreter does not care:

```
$ python3           
Python 3.6.8 (default, Nov 16 2020, 16:55:22) 
[GCC 4.8.5 20150623 (Red Hat 4.8.5-44)] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> #!/bin/python3
... 
>>> import openstack
>>> 
```

It is important to put the shebang to mark if python2 or 3 should be used. Now it works with both, but that may not be the case in the future.

docs/cloud/pouta/api-access.md
lvarin: Ok

docs/cloud/pouta/api-access.md
tdeneke: Ok. The password is now only under `secure.yaml`. Hopefully users do not commit that file by mistake too :).

lvarin: Thanks, they will probably do it anyways :smile:

docs/cloud/pouta/api-access.md
```diff
@@ -0,0 +1,141 @@
...
+clouds:
+  openstack:
+    auth:
+      password: XXXXXXXXXX
+```
+
+Now, you can run the following simple example which lists the available virtual machine flavors:
+
+```
+#!/bin/python3
```
lvarin, line 125:
```suggestion
#!/usr/bin/python3
```

lvarin (approved):

PR #907, opened 2021-04-01
Restructure computing/containers pages

mvsjober:
Preview can be viewed here: https://csc-guide-preview.rahtiapp.fi/origin/new-singularity/computing/containers/run-existing/


jessepharrison (approved):
I think this looks very nice. This part of the documentation is probably good to keep general + succinct, but we could link to a separate tutorial with some details / practical tips for building Singularity files from scratch.

PR #920, opened 2021-04-12
Connecting with SSH keys, reorganizing + RStudio/Jupyter Notebook tutorial

ktiits:


jussienko:
Lots of good stuff, however, I feel that the tunneling related stuff is here out of context (i.e. note that powershell does not support jump hosts and instructions for setting tunneling for Putty). I would suggest to discuss tunneling settings in the context of relevant application (i.e. RStudio or Jupyter), or if it feels that there are generic enough use cases we should have a separate "ssh tunneling artcile".

jussienko (commented):
Excellet work!

Tiny comments: 

"Manual copying keys" part has in the text Windows syntax for path (`\`) but Linux one (`/`) in the example command.

Link to tutorial is missing from the main tutorial page (support/tutorials/).

Maybe the link to tutorial could be included in the first section of "interactive-usage", and the two examples later on completely removed.

ktiits:
> "Manual copying keys" part has in the text Windows syntax for path (\) but Linux one (/) in the example command.
Only the Windows example left with \, others changed to /.

> Link to tutorial is missing from the main tutorial page (support/tutorials/).
Added.

Maybe the link to tutorial could be included in the first section of "interactive-usage", and the two examples later on completely removed.
> Link added and 2 examples removed.

jussienko (approved):
Mycket bra!

PR #941, opened 2021-05-11
Missing host information added

fischuu:
In Paragraph 3 of the Nextcloud part of the manual is the host information missing, maybe it is obvious what needs to be written there, but it took me some time to figure out what it is.


Nortamo:
Fixed in  #947

PR #946, opened 2021-05-17
start an ad-hoc pod with non default command

tdeneke:
- Update the instruction to include how to start an ad-hoc pod with non default command specified under the container image.


lvarin (approved):
LGTM

PR #950, opened 2021-05-26
Add warning about high number of files in volumes

lvarin:
Add warning about high number of files in volumes and the time it takes to mount


tdeneke (approved):
\lgtm

PR #955, opened 2021-05-28
Remark in docs that Pods diying is expected

lvarin:


tdeneke (commented):
a bit more explicit wording.

tdeneke (approved):

PR #959, opened 2021-06-04
Link csc env eff

attesillanpaa:
added links to using csc hpc environment efficiently to training materials and tutorials lists


jussienko (approved):
Good!

PR #984, opened 2021-06-22
Csccva nsys ncompute 01

csccva:
Added `nsys` to the list and did some minor changes to the `nvprof` page.


jussienko (approved):
Nice work

PR #987, opened 2021-06-29
RAM requirement for standard.large is wrong

khabiruddin:
In standard.large flavor, amount of ram needed is 8GB but 7GB is
written instead.

Reviewer can check it by running command,
"openstack flavor show standard.large" and then check for ram.
Then compare it with https://docs.csc.fi/cloud/pouta/vm-flavors-and-billing/#standard-flavors.


junousi:
See details on jira.ci , closing.

PR #1010, opened 2021-07-16
Updated amd documentation to cover 2021

nruneberg:
https://csc-guide-preview.rahtiapp.fi/origin/ams_update/apps/ams/


MatiasJJ (changes requested):
In the first note the "fast local storage on Puhti" is mentioned. Is that NVMe? It could include "NVMe" to clarify and to distinguish it from login node's local scratch.

MatiasJJ (approved):

PR #1019, opened 2021-08-10
Updated Molpro to cover version 2021.2

nruneberg:
... and added a small example on performance
https://csc-guide-preview.rahtiapp.fi/origin/molpro_update/apps/molpro/


galfthan:
Is this just bugfix update, or a point release that we should mention in whats new?

attesillanpaa:
Version bumped from 2019.2 to 2021.2. I guess it's a major update. Unless instructed otherwise, I'd go for whatsnew. BTW, there's a cp2k update appearing there, too. Better to keep these separate or combine?

attesillanpaa (approved):

PR #1034, opened 2021-08-27
Update how-to-create-new-user-account.md

samisalonen-csc:
added minor changes to wording, we need organizational email from non-haka users


kkmattil (approved):
OK

PR #1036, opened 2021-08-30
Make Rahti's storage page more similar to Pouta's.

lvarin:
I made Rahti's storage page more similar to Pouta's
https://docs.csc.fi/cloud/pouta/storage/

Also I added a drawing


docs/cloud/rahti/storage/index.md
```diff
@@ -1,9 +1,25 @@
...
+## Persistent storage
+
+The solution more similar to Docker volumes is a [Persistent volume](/cloud/rahti/storage/persistent/). This is a volume located in an external storage, that is mounted directly into the filesystem of the container, like an additional hard disk would. This is the best solution for a seamless storage solution. You can also un-mount it from the Pods it is attached and attach it to another set of Pods. Moreover, persistent volumes represent a reliable storage option as the data in them is replicated.
+
+You can find additional information on the [Persistent volume](/cloud/rahti/storage/persistent/) page.
+
+## Ephemeral storage
+
+Other solution is to use ephemeral storage, called in Kubernetes an _Empty dir_. This is only meant for storing intermediate or temporal data, that needs fast read write access by the applications running inside the containers. The same _Empty dir_ can be mounted in every container inside a Pod. The data in the ephemeral storage is lost when the Pod is deleted. Given that the data in the ephemeral you should not store any persistent data there, and the ephemeral storage should not be considered reliable.
```
tdeneke, line 17:
```suggestion
Another solution is to use ephemeral storage, also known in Kubernetes as an _Empty dir_. This is only meant for storing intermediate or temporal data that needs fast read-write access by applications running inside containers. The same _Empty dir_ can be mounted in every container inside a Pod. The data in the ephemeral storage is lost when the Pod is deleted and as such, should not be used to store any persistent data. The purpose of the ephemeral storage framework is to facilitate the request for guaranteed local storage mainly by enabling pods to specify their temporal local storage needs. 
```

docs/cloud/rahti/storage/index.md
```diff
@@ -1,9 +1,25 @@
...
+
+You can find additional information on the [Ephemeral storage](/cloud/rahti/storage/ephemeral/) page.
+
+## Object storage
+
+In case you need to store large volumes of data, or you need that your data can be easily accessed over the Internet, e.g., using URLs, object storage is the storage solution you are looking for. By using object storage, you are free to create and delete your Pods as the object storage is not tied to any Pod. Moreover, your data is replicated, thus object storage represent a reliable storage option.
```
tdeneke, line 23:
```suggestion
In case you need to store large volumes of data, or you need that your data can be easily accessed over the Internet, e.g., using URLs, object storage is the storage solution you are looking for. By using object storage, you are free to create and delete your Pods as the object storage is not tied to any Pod. Moreover, your data is replicated, thus object storage represents a reliable long-term storage option.
```

tdeneke (commented):
Would add value if texts in the image included are made bigger. Also consider a lighter image background.

tdeneke (approved):
looks good.

PR #1042, opened 2021-09-01
Sensitive data update

fmorelloCSC:
updated SD Desktop with new instructions regarding Desktop naming / Additional disk space /Integrated videos in text


kkmattil (approved):
In the case of SD Desktop, we end up to a bit confusing terminology:
Here "SD Desktop" refer often to virtual machine, while in many other occasions "Desktop" refers to the graphical working environment. For example:
"Thus, if you add colleagues/collaborators as project members in the same CSC project, they will be able to access the same Desktop."  Here this means that they can access the same virtual machine, but the actual "Linux desktop" where you have your applications running is not shared.

PR #1055, opened 2021-09-15
WIP: add other force version available

samumantha:
* 3.5.1 seems to be the default version atm

@ktiits : Should default be updated to 3.6.5 or documentation show that 3.5.1 is default (and for what reason: is the finnhub issue the reason?)?


samumantha:
Update: newer version is default with singularity wrapper, older version is not wrapped and therefore would need different use instructions which may be too confusing. Closing this.

PR #1089, opened 2021-11-08
Added/updated OOD usage to GIS modules with GUI

samumantha:
WIP because of NoMachine way.

Should all NoMachine instructions be removed from these pages, @ktiits ?

For SAGA and GRASS GIS the Nomachine instructions are combined with interactive usage. Should the interactive usage part be kept or completely removed?

### DIFF
--- a/docs/apps/grass.md
+++ b/docs/apps/grass.md
@@ -10,33 +10,23 @@ __GRASS__ is available in Puhti with following versions:
 
 ## Usage
 
-### Using GRASS GIS Command Line Interface 
+### GRASS GIS Command Line Interface 
 
-In Puhti, GRASS GIS is installed inside the [QGIS singularity installation](qgis.md)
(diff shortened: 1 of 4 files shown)


ktiits:
> WIP because of NoMachine way.
> 
> Should all NoMachine instructions be removed from these pages, @ktiits ?
> 
> For SAGA and GRASS GIS the Nomachine instructions are combined with interactive usage. Should the interactive usage part be kept or completely removed?

I would remove everything related to NoMachine, but leave the commands visible that are required for launching the GUI from terminal.

For Saga, GRASS and SNAP, I would say that OOD is recommended way for GUI, but option 1 for computing would still be CLI. 

Its likely time to remove the older saga-gis documentation from docs.

samumantha:
Should the whole old saga-gis module be removed, or only usage instructions?

samumantha:
SAGA and GRASS have only instructions for interactive use, should this be rephrased, to include also usage from batch scripts?

ktiits:
> SAGA and GRASS have only instructions for interactive use, should this be rephrased, to include also usage from batch scripts?

Yes. GRASS has examples in Github. For SAGA Rsagacmd could be recommended for scripting.

ktiits (approved):

PR #1094, opened 2021-11-10
About links

mlehtiva:
Some notes about links


jussienko (changes requested):
Could you double check that changes I made look fine?

jussienko (approved):
Thanks for checking!

PR #1096, opened 2021-11-12
Fix images

Nortamo:
Github Action to fix drag and drop images


Nortamo:
#1097 Is the proper PR for this

PR #1106, opened 2021-11-23
QGIS version update

samumantha:
* updated version to 3.22 -> new default QGIS
* removed module load instructions, since it does not actually work that way in Puhti web interface (needs host terminal?)
* updated availability of GRASS (7.8.5) and also removed the module load instructions there (same as above) for Puhti web interface

### DIFF
--- a/docs/apps/grass.md
+++ b/docs/apps/grass.md
@@ -22,7 +22,12 @@ Using GRASS GIS in [Puhti web interface with Desktop app](../computing/webinterf
 1. Log in to [Puhti web interface](https://puhti.csc.fi). 
 2. Start GRASS GIS with Apps -> Desktop, choose Desktop: 'None' and App: 'GRASS GIS'.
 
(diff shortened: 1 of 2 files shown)


samumantha:
WIP because both QGIS and GRASS should keep the module load commands, but they do not work on OOD (currently only when started from host terminal), so should be put separately from OOD instructions.

-> Should NoMachine instructions be kept for that or is OOD supposed to be replacing NoMachine in the future (ie Users should have access to the host terminal?)? Maybe @galfthan has some insights? 

-> Does X connection automagically work for all Unix users (When I (CentOS) start GUI on puhti it magically opens on my desktop without any extra keywords, is that the case for everyone, if not what needs to be done to make this possible ?)?

samumantha:
Put back terminal commands after discussion with @ktiits 

-> new OOD (coming ~ next week) will have host terminal where commands will work.

Docs then still need adaption to make things clearer!

ktiits (approved):

PR #1132, opened 2021-12-20
Update disk billing information to 2022

MatiasJJ:
NOTE: Should not merge before 2022 when these policies apply.
This is made according to the information email sent on 09.12.21.


MatiasJJ:
I use this information to update EFF-material so I'd like to know it's accurate.

galfthan (approved):
Fixed text, this is now ok

PR #1141, opened 2022-01-18
New module: Sen2mosaic

samumantha:
User request: RT#523933

* Sen2mosaic is a research software for downloading, preprocessing and mosaicing Sentinel-2 satellite remote sensing data.
* Original command line calls had spaces in them, which have been replaced with `_` (therefore the note in the guide)
* `s2m_preprocess` is a weird implementation of sen2cors (https://docs.csc.fi/apps/sen2cor/) `L2A_Process` and can be configured using `L2A_GIPP.xml` file that is provided with sen2cor (if the user does not want to configure, then the default GIPP file from sen2cor installation is used, see also (https://github.com/CSCfi/csc-env-guide/pull/328). 

* similar **advanced usage** note was also added to sen2cor module


samumantha:
Screenshot 2022-01-18 at 10-31-00 Sen2mosaic - Docs CSC

ktiits (approved):

PR #1143, opened 2022-01-19
scrap support menu level

attesillanpaa:
FAQ and tutorials were too hard to find. They are made more visible by removing the "support" menu layer. See https://csc-guide-preview.rahtiapp.fi/origin/flatten-menu/


rkronberg (approved):
Good improvement!

PR #1149, opened 2022-01-25
Create new build from Dockerfile

lvarin:
Give another example on how to build an image


tdeneke (approved):
`--name ` parameter could be added (optional)

PR #1156, opened 2022-02-03
Switch tool for checking public IP to csc tool

matteopozza:
When setting up a virtual machine in cPouta, it can be useful to know
the user's IP address so that the security rules for the virtual
machines can be tailored accordingly. At the moment, the tool that the
documentation is recommending for checking one's own IP address is a
third-party tool.
CSC offer its own tool to find out what is the public IP address of the
device accessing it. This commit changes the third-party tool to the
CSC-offered one.


jarnolaitinen (approved):
OK

Aegil (approved):

PR #1164, opened 2022-02-11
Parallel R tutorial

jessepharrison:
The idea here was to write up some generic information on R parallelisation strategies (mostly for users who are unfamiliar with this topic). There are no code examples as such, but the tutorial refers to existing examples on the `r-env-singularity` page. Many of the concepts aren't truly R-specific, but might be easier for R users to digest if we put them under this header.

A separate "code-heavy" parallel R document could also be considered in the future, if we want to have more examples available.

Preview:
https://csc-guide-preview.rahtiapp.fi/origin/parallel-r-tutorial/support/tutorials/parallel-r/


jussienko (changes requested):
I am not familiar how R handles parallelization in practice, but I find the discussion about "multicore" and "multithread" analysis very confusing.

A "core" is the physical hardware component doing the work.

A "thread" is a software construct at least in standard terminology (hardware thread is sometimes used in the context of simultaneous multithreading, but that it is not enabled in Puhti).

I wonder if "multicore" means actually "multiprocess"?

I think the sentence "Certain R packages support multithreading, which makes it possible to parallelise processes taking place within an individual core." also a bit problematic, only in a very rare circumstances (I/O bottleneck) one should parallelize within a **core**.

jussienko (approved):
Huge improvement, tutorial looks now very good!

PR #1168, opened 2022-02-15
SGID documentation

samiilvonen:
Clarification on the SGID settings on the project and scratch folders.


samiilvonen:
Added a short note on ls command output and correct permissions.

PR #1178, opened 2022-02-24
New geoconda added

samumantha:
* new geoconda 3.9.10
* with one new package (cfgrib)
* 2 new references (by @ktiits )


ktiits:
* The packages are otherwise in alphabetic order, so please move cfgrip to right place
* Under "Additionally geoconda includes:" needs likely also updates
* I guess 3.9.10 has also wrappers? In that case I would keep the descriptions similar. 
* 3.9.10 (Singularity installation, for normal use-case, this should not have an influence on usage)
* 3.8.8 (Singularity installation with wrappers, for normal use-case, this should not have an influence on usage)

samumantha:
-> 'Under "Additionally geoconda includes:" needs likely also updates'
Did you have anything specific in mind that was missing here @ktiits ?

samumantha:
WIP because waiting for 2 more updates to be added as well (currently new module is not default yet, so should not hurt)

ktiits (approved):

PR #1193, opened 2022-03-10
Doc: Find hostname of a floating IP

feber:
Additional information for cPouta users, on how they can find hostname of a floating IP.

Edit: [link to section](https://docs.csc.fi/cloud/pouta/additional-services/#dns-services-in-cpouta).

### DIFF
--- a/docs/cloud/pouta/additional-services.md
+++ b/docs/cloud/pouta/additional-services.md
@@ -44,7 +44,7 @@ All our floating IPs are mapped to a default hostname, for example:
 vm0120.kaj.pouta.csc.fi
 ```
 
-To find hostname of a floating IP, you could leverage `host` command:
+To find the hostname of a floating IP, you can use `host` command:
 ```sh
 host -a <floating IP address>
 ```
@@ -52,7 +52,7 @@ host -a <floating IP address>
 And the result is `vmXXXX.kaj.pouta.csc.fi` listed in output:
 
 ```
-... omitted for brevity
+...
 ;; ANSWER SECTION:
 x.x.x.x.in-addr.arpa. xxx IN     PTR     vmXXXX.kaj.pouta.csc.fi.
 ```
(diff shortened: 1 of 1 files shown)


docs/cloud/pouta/additional-services.md
```diff
@@ -38,12 +38,25 @@ the CSC Service Desk to coordinate this.
...
 cPouta does not currently offer integrated name service management.
 
-All our floating IPs are mapped to a default hostname, for example
+All our floating IPs are mapped to a default hostname, for example:
 
 ```
 vm0120.kaj.pouta.csc.fi
 ```
 
+To find hostname of a floating IP, you could leverage `host` command:
```
matteopozza, line 47: You can add "the" in front of "hostname" -> "To find _the_ hostname of a floating IP..."

I would also rephrase "you could leverage host command" to a simpler "you can use the host command" or "you can leverage the host command"

docs/cloud/pouta/additional-services.md
```diff
@@ -38,12 +38,25 @@ the CSC Service Desk to coordinate this.
...
+```sh
+host -a <floating IP address>
+```
+
+And the result is `vmXXXX.kaj.pouta.csc.fi` listed in output:
+
+```
+... omitted for brevity
```
matteopozza, line 55: I think you can remove this line. Alternatively, you can just keep the three dots, i.e., without the "omitted for brevity". In my opinion, the three dots already indicate that something else is expected before the subsequent lines, so they are sufficient.

docs/cloud/pouta/additional-services.md
feber: addressed, thanks

docs/cloud/pouta/additional-services.md
feber: fair enough. yes I agree that `omitted for brevity` here is kind of duplicate. thanks.

matteopozza (approved):
This is a nice contribution to our documentation, thanks!

PR #1195, opened 2022-03-11
JupyterHub and Notebook custom image how-tos

tdeneke:
- Added pointer to JupyterHub and Notebook image customization how-tos


lvarin (approved):
LGTM

PR #1196, opened 2022-03-11
Add note on science-specific support, link to research.csc.fi

rkronberg:
https://csc-guide-preview.rahtiapp.fi/origin/sciences-help/apps/


nruneberg (approved):
Ok

PR #1209, opened 2022-03-16
Change to rockylinux as centos reached its EOL

lvarin:
Centos8 EOL was back in December 2021. Rocky Linux was recommended by Teemu.


tdeneke (approved):
looks a solid alternative.

PR #1211, opened 2022-03-16
Migrate accessibility statement from www.csc.fi to docs.csc.fi

rkronberg:
https://csc-guide-preview.rahtiapp.fi/origin/migrate-accessibility-statement/support/accessibility/


rkronberg:
The absolute internal link to the accessibility statement in the footer does not work in the preview, but works in a local mkdocs build of the website and should work in docs.csc.fi.

attesillanpaa (approved):

PR #1215, opened 2022-03-18
Lumi linking

rkronberg:
https://csc-guide-preview.rahtiapp.fi/origin/lumi-linking/


rkronberg:
I think the first iteration could be published to get something online. Later it would be good to add information/links concerning what kind of workloads are most suitable for LUMI and perhaps links from all Docs CSC software pages to LUMI if there exists an optimized EasyBuild recipe (or pre-installation).

PR #1223, opened 2022-03-23
Rockylinux and new nginx

lvarin:
When activating in Rahti the new rocky linux build, some changes were necessary:

* Add the RPM findutils for the xargs command.
* Enable the nginx mainline module


tdeneke (approved):
looks good.

PR #1231, opened 2022-03-31
make some links more accessible

attesillanpaa:
formulate as link, add a screen reader friendly text

https://csc-guide-preview.rahtiapp.fi/origin/tykky-links/computing/containers/tykky/


rkronberg:
Should Tykky have an entry under "Apps"? And perhaps add to what's new?

attesillanpaa:
What's new: absolutely! I guess Apps entry would be ok, too.

rkronberg (approved):
Ok. Leaving out the apps entry for now, would require moving tykky.md under apps or then editing the scripts.

attesillanpaa:
Maybe we can (later) make a stub entry in apps, and point here from there?

PR #1237, opened 2022-04-01
Replace nomachine with ood

rkronberg:
Recommend puhti web interface instead of nomachine in several places in docs.


galfthan:
Based on a quick scan of the changes it looks good. I assume now nomachine is only mentioned in its own tutorial.

rkronberg:
That's correct. I also left the Paraview and Visit apps-pages untouched because I'm unsure whether all the currently documented features would work in the web interface, e.g. running paraview on GPUs

attesillanpaa (approved):

PR #1246, opened 2022-04-06
Fix partitions 2

galfthan:
- Updated puhtis interactive limits to be correct (MaxTRESPU       cpu=8,gres/nvme=720,mem=76G)
- Tuned table on mahtis interactive partition and reorganized text a bit


rkronberg (approved):
Ok

PR #1263, opened 2022-05-02
Update fireworks installation instructions with tykky

rkronberg:
https://csc-guide-preview.rahtiapp.fi/origin/fireworks-tykky/computing/running/fireworks/


nruneberg (approved):
Good! Despite some redundancy in the documentation I think that it's now more self-contained and aids in proceeding to the solution.

PR #1264, opened 2022-05-02
Emphasize the lack of automatic backups

rkronberg:
https://csc-guide-preview.rahtiapp.fi/origin/no-backups/computing/disk/


CSC-swesters (approved):
Content looks good to me.
Prefer to keep markdown lines at most 100-120 columns wide in the future, so that it is easier to read :) Tables cannot be wrapped of course, but normal paragraphs can, e.g. line 10.

PR #1283, opened 2022-05-13
Root cleanup

joonas-somero:
Removed some unused and misplaced files from repository root.


rkronberg (changes requested):
Nice! I noticed that there's links to the removed changelog and todo files in `README.md`. I think we should remove those too :)

rkronberg (approved):

PR #1284, opened 2022-05-16
WIP: Add level 1 headers to TOC to fix search function

joonas-somero:
Search seems to work now. Hadn't generated content for 'What's new' when testing locally, so didn't notice TOC change on landing page.

https://csc-guide-preview.rahtiapp.fi/origin/lower_toc_depth_bottom/

### DIFF
--- a/mkdocs.yml
+++ b/mkdocs.yml
@@ -229,7 +229,7 @@ extra_javascript:
 
 markdown_extensions:
   - toc:
-      toc_depth: 2-6
+      toc_depth: 1-6
   - admonition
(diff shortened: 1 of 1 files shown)


Nortamo:
Looks good to me, adding two images here to illustrate the effect. 
As a clarification, the search was broken in the way that it did not serve the content directly under the level one main header instead the first match from a level 2 header -> main summary of a lot of pages not showing up. 

## Old

docs1

## New

docs2

Nortamo (approved):

PR #1286, opened 2022-05-18
Check licenses

joonas-somero:
A test that every application page has a level 2 or 3 heading starting with the capitalized word 'License'. The check for the trailing 's' is redundant at the moment since the regex doesn't match end-of-line after '##(#) License(s)' anyway. For example "## License and acknowledgement" passes the test.

https://csc-guide-preview.rahtiapp.fi/origin/check-licenses/

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -45,3 +45,6 @@ script:
   - bash tests/check_fixme.sh
   # Test that no files are masked by folder/index.md
   - bash tests/check_masked.sh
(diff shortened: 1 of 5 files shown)


joonas-somero:
As expected, the build fails and the next step is to implement a whitelist system.

attesillanpaa:
For now, adding the missing License headers to the few failing apps sounds good.

attesillanpaa (approved):
ok!

PR #1303, opened 2022-06-02
Delete pktools.md

ktiits:
pktools is no longer maintened... Only 1 user in 1,5 years.


samumantha (approved):
Sorry for the wait, I thought I pushed that button already

PR #1309, opened 2022-06-06
Add throughput whatsnew

rkronberg:
https://csc-guide-preview.rahtiapp.fi/origin/throughput-wn-fixes/support/wn/training-new/#documentation-on-high-throughput-computing-and-workflows-updated-662022


rkronberg:
The unstyled content (icons and header links) in the workflow matrix might be related to firefox being buggy. The content looks fine with chromium, and after waiting a while also firefox seems to format it correctly. I'd say no action is needed at the moment, but if the problem arises again and persists we should look into it in more detail.

attesillanpaa (approved):

PR #1321, opened 2022-06-09
Example with running nextflow with HQ

Nortamo:
Placement TBD. 
I.E should it be an app page, tutorial or something else

### DIFF
--- a/docs/support/tutorials/nextflow.md
+++ b/docs/support/tutorials/nextflow.md
@@ -1,0 +1,59 @@
+# High throughput nextflow workflow using hyperqueue
+
+
+## Example batch script
+```
+#!/bin/bash
+#SBATCH --partition medium
+#SBATCH --account project_2001659 
+#SBATCH -N 3
+#SBATCH --exclusive
+#SBATCH --time=00:10:00
+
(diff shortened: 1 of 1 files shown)


rkronberg:
https://csc-guide-preview.rahtiapp.fi/origin/hq_nf/support/tutorials/nextflow-hq/
Added a short motivation and links. I think this could be published now.

rkronberg:
Regarding whether it should be an app page or tutorial, I think tutorial is suitable in this case. That said, it might be worthwhile to create separate app pages for Nextflow and HyperQueue with some very basic instructions/CSC-specific details.

Right now we have two tutorial pages for Nextflow which are more or less bioinformatics- and Puhti-specfic. These could be reworked/generalized at some point.

jlento (changes requested):
In the batch script, one could first assign the per job work dir with something like

# Per job work dir
wrkdir=$PWD/WRKDIR-$SLURM_JOB_ID

and then later in the script use the variable where appropriate.

Second, where does file `main.nf` come from (line 48)?

jlento (approved):

PR #1326, opened 2022-06-10
Implemented preliminary breadcrumbs navigation

joonas-somero:
https://csc-guide-preview.rahtiapp.fi/origin/breadcrumbs/

Breadcrumbs navigation as text only (not as links) at the moment. The feature in question pictured here above the heading "Introduction":

breadcrumbs


rkronberg (approved):
Ok, looks good! At some point it would be good to extend the breadcrumbs also to FAQ/tutorials categories as many pages link to FAQs/tutorials and following these links may be quite a common step where users reading docs get "lost"

PR #1328, opened 2022-06-13
Add Jupyter Dask guidelines

ktiits:


docs/support/tutorials/dask-python.md
```diff
@@ -114,8 +114,26 @@ for dataset in datasets:
...
 
-### References 
+## Dask with Jupyter
+
+For better understanding of Dask computing exists [Dask JupyterLab Extension](https://github.com/dask/dask-labextension). In Puhti it is currently available only in [geoconda](../../apps/geoconda.md) module, if you need it together with other Python modules, ask from Servicedesk.
+
+To use Dask JupyterLab Extension:
+
+
+* [Start JupyterLab session](../../computing/webinterface/jupyter.md) in Puhti web interface. Select `geoconda` as module. Reserve computing resources for you Local cluster, notice the [interactive job](../../computing/running/interactive-usage.md) limits. Bigger requests are sent to usual queueing system. Max. 40 cores.
```
robinkar, line 124:
```suggestion
* [Start JupyterLab session](../../computing/webinterface/jupyter.md) in Puhti web interface. Select `geoconda` as module. Reserve computing resources for your Local cluster, notice the [interactive job](../../computing/running/interactive-usage.md) limits. Bigger requests are sent to usual queueing system. Max. 40 cores.
```
Small typo

robinkar (changes requested):
Looks good.

ktiits:
I pretty much re-wrote now everything. In my tests the new Cluster button nicely Dask panels open, but they show nothing.. I wonder if i have messed up some setting, but even when I deleted labextension.yaml and let Dask create a new one, situation did not change..

JupyterLab Dask Extension works for me only with SSH tunnelling. 

But the basic Dask Dashboard works, if you now what URL to use, so added that now to documentation. Dask Dashboard is available also with python-data.

On the good side I found how to fix SlurmCluster problems with manually defining Python path.

ktiits:
We tested today with Samantha's account too (she has not used Dask before, nor modified the Dask setting yml files) and the same result. The Dask extension did not work with OoD. The Dask Dashboard worked ok, if opened in separate web-browser tab.

robinkar (approved):
Looks good.
The current instructions work fine. Though going through `https://puhti.csc.fi/rnode` seems to break the Info tab in the dashboard. So SSH tunneling for the dashboard port could also be good alternative, depending on how commonly used the Info tab is. 

The Dask JupyterLab extension seems to work fine for me with geoconda module, but I might be using Dask in some unusual way as I'm not very familiar with Dask.
For example
```
from dask.distributed import Client

client = Client("tcp://127.0.0.1:40215")

import dask.array as da
x = da.random.random((10000, 10000), chunks="16 MiB")
y = (x+x.T) -x.mean(axis=0)
y.sum().compute()
```
show the graphs fine. `Client("...")` is the scheduler address from creating a new cluster using the extension (the client code can be autogenerated by clicking the `<>` button on the cluster).

PR #1336, opened 2022-06-20
WIP: CSC Design System implementation

joonas-somero:
Incrementally implementing the CSC Design System for Docs. Very rough at the moment.

https://csc-guide-preview.rahtiapp.fi/origin/preliminary-design-system/


Nortamo:
Minor thing and does not need to be changed, but the contrast for highlighted text on the background seems a bit weak. 
screen

joonas-somero:
> Minor thing and does not need to be changed, but the contrast for highlighted text on the background seems a bit weak. screen

Yes, there are many artefacts left from the current style that need to be changed. I threw something together to start with and will proceed incrementally from here. Any input is highly appreciated, though.

PR #1346, opened 2022-06-29
Remove outdated/unnecessary Allas pages

rkronberg:
https://csc-guide-preview.rahtiapp.fi/origin/remove-old-allas/data/Allas/


oscarkraemer (approved):
I feel that I should not be allowed to approve this, but anyhow. I think removing the content is a good idea because I think it have served it purpose of explaining the changes of `object.pouta.csc.fi` to `a3s.fi` .

PR #1347, opened 2022-06-30
Cp2k update

rkronberg:
https://csc-guide-preview.rahtiapp.fi/origin/cp2k-update/apps/cp2k/

Edit note on elpa (and split lines for readability). There were some speed (and memory) issues that we speculated to be related to elpa (ergo the previous disclaimer), but these were actually tracked down to the openblas library. Switching to amdblis seems to restore the expected performance improvements of elpa and also fixed the issue of calculations using up to 7 times more memory than previously (on rhel7). Updated versions to be installed on Mahti soon.


nruneberg (approved):
Ok

PR #1359, opened 2022-07-11
Update R part to terra

ktiits:


samumantha (approved):
If `terra` is preferred over `raster`; why keep `raster` example here?

ktiits:
raster is still very widely used, so left it for now, could be deleted after some years.

PR #1377, opened 2022-08-22
Monospaced text inside headings

joonas-somero:
Monospaced text inside headings was rendering as it renders inside text blocks. These changes fix that and add the case in question to the reference card. Here's a page with a heading containing monospaced text:

https://csc-guide-preview.rahtiapp.fi/origin/monospace-in-headings/cloud/rahti/tutorials/connect-database-hpc/#option-2-using-the-oc-command-line-tool


rkronberg (approved):
LGTM

PR #1378, opened 2022-08-22
Update orca.md

nruneberg:
https://csc-guide-preview.rahtiapp.fi/origin/orca_localdisk/
Emphasizing more clearly the benefits of using local disk for wft based correlations methods


rkronberg (approved):
Ok

PR #1384, opened 2022-08-24
Updated the info about Turbomole

mariavd:
Updated the version number and the reference. Added a link to the tutorial for beginners, and a general tip to users. Small style corrections.


nruneberg (approved):
Good! Maybe the general recommendation to prefer SMP over MPI has to be checked in more detail

PR #1391, opened 2022-09-01
Applications by license

joonas-somero:
https://csc-guide-preview.rahtiapp.fi/origin/license-tags/apps/by_license/

image


Nortamo:
Should we link/add some information on what exactly the different types mean or is this directly apparent to the user?

joonas-somero:
> Should we link/add some information on what exactly the different types mean or is this directly apparent to the user?

Another option would be to have the tags be just a bit longer. Like 'Free for academic use'.

rkronberg:
Perhaps there could be a short description under each tag on the index page? And clicking the tags on the application page would link to the corresponding anchor on the index page with the explanation?

joonas-somero:
> This looks very nice! Would it make sense to add a what's-new item e.g. in the applications category: (something along the lines) new view to available software: grouped by license type.

That _would_ make sense, yes. I'll just add that before merging.

rkronberg:
Looks good! I guess this can be merged?

PR #1394, opened 2022-09-02
Hq instructions

Nortamo:
Basic instructions for using hyperqueue
https://csc-guide-preview.rahtiapp.fi/origin/hq_instructions/apps/hyperqueue/

### DIFF
--- a/docs/apps/hyperqueue.md
+++ b/docs/apps/hyperqueue.md
@@ -76,9 +76,9 @@ the workers early.
 Loop until all workers are online (note no timeout):
 
 ```bash
-num_up=$(hq worker list | grep -c RUNNING)
 while true; do
 
+    num_up=$(hq worker list | grep -c RUNNING)
     echo "Checking if workers have started"
     if [[ $num_up -eq $SLURM_NNODES ]]; then
         echo "Workers started"
@@ -86,7 +86,6 @@ while true; do
     fi
     echo "$num_up/$SLURM_NNODES workers have started"
     sleep 1
-    num_up=$(hq worker list | grep RUNNING | wc -l)
 
 done
 ```
--- a/docs/apps/hyperqueue.md
+++ b/docs/apps/hyperqueue.md
@@ -16,8 +16,8 @@ Free to use and open source under [MIT License](https://github.com/It4innovation
 
 ## Available
 
-* Puhti: 0.9.0
-* Mahti: 0.11.0
+* Puhti: 0.4.0, 0.9.0
+* Mahti: 0.11-dev
 
 ## Usage
 
@@ -218,7 +218,7 @@ the system batch scheduler.
 Using Snakemake's `--cluster` flag we can use `hq submit` instead of `sbatch`:
 
 ```bash
-snakemake --cluster "hq submit --cpus {threads} .."
+snakemake --cluster "hq submit --cpus <threads> ..."
 ```
 
 If you are porting a more complicated workflow from Slurm, you can do
@@ -233,8 +233,10 @@ how to use HyperQueue as an executor for Nextflow workflows.
 ## More information
 
 !!!info "MPI"
-    Although HyperQueue does not do MPI execution out of the box, it's possible using a
-    combination of the HQ feature "Multinode Tasks" and `orterun`, `hydra` or `prrte`.
-    This way one can schedule MPI tasks at a node-level granularity.
+    Although HyperQueue does not do MPI execution out of the box, it's possible
+    using a combination of the HQ feature [Multinode
+    Tasks](https://it4innovations.github.io/hyperqueue/stable/jobs/multinode/)
+    and `orterun`, `hydra` or `prrte`. This way one can schedule MPI tasks at a
+    node-level granularity.
 
 * [Full documentation for HQ](https://it4innovations.github.io/hyperqueue/v0.11.0/)
--- a/docs/apps/hyperqueue.md
+++ b/docs/apps/hyperqueue.md
@@ -110,11 +110,11 @@ hq submit --array 1-10 --cpus <n> <COMMAND>
 `<COMMAND>` then has access to the environment variable `HQ_TASK_ID` which is used
 to enumerate all the tasks.
 
-!!! info "sbatch-hq (not yet available)"
-    For very simple submissions where you only want to run each line within a file with identical
-    resources (task farming) you can soon just use the CSC utility tool `sbatch-hq`. This way you
-    do not have to care about HyperQueue. Once available, run `module load sbatch-hq` to load the
-    wrapper.
+!!! info "sbatch-hq"
+    For very simple submissions where you only want to run each line within a file with
+    identical resources (task farming) you can just use the CSC utility tool `sbatch-hq`.
+    This way you do not have to care about HyperQueue. Run `module load sbatch-hq` to load
+    the wrapper (only available on Mahti).
 
 When we have submitted everything we want, we need to wait for the jobs to finish.
 This can be done e.g. with:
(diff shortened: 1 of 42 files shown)


j-kali:
I can see that you're recommendingI can see that you're recommending to run the server inside the job, which is great. But one has to be careful when starting workers immediately after, as it takes a moment for the server to start and if you start workers too early they might fail immediately.

At least that was my experience (couple of versions ago though). In any case it might cause a race condition which gives rather poor user experience. My solution at the time was to wait for the server to be up with a simple `until` loop. To paste a piece of my old code

```
# absolutely horrible but for now I don't know a better way of waiting for the server to be actually there
until ${hq} jobs "--server-dir=${hq_server_dir}" ; do sleep 1 ; done
```

Nortamo:
> I can see that you're recommendingI can see that you're recommending to run the server inside the job, which is great. But one has to be careful when starting workers immediately after, as it takes a moment for the server to start and if you start workers too early they might fail immediately.
> 
> At least that was my experience (couple of versions ago though). In any case it might cause a race condition which gives rather poor user experience. My solution at the time was to wait for the server to be up with a simple `until` loop. To paste a piece of my old code
> 
> ```
> # absolutely horrible but for now I don't know a better way of waiting for the server to be actually there
> until ${hq} jobs "--server-dir=${hq_server_dir}" ; do sleep 1 ; done
> ```

Ah good point, I have not had issues with that, but it's easy to see that some temporary lag spike might lead to the server not being up when the workers are started. I think that loop is probably fine, no worse then the one where I wait for the workers to come online, I'll add it.

docs/apps/hyperqueue.md
```diff
@@ -0,0 +1,229 @@
...
+#SBATCH --time=01:00:00
+
+export SLURM_EXACT=1
+
+# Load the required modules
+module load hyperqueue
+
+# Set the directory which hyperqueue will use 
+export HQ_SERVER_DIR=$PWD/hq-server-$SLURM_JOB_ID
+mkdir -p $HQ_SERVER_DIR
```
j-kali, line 146: `mkdir -p "$HQ_SERVER_DIR"`

`$PWD` might have spaces, it doesn't expand in assignments but here it could so quotation needed

docs/apps/hyperqueue.md
```diff
@@ -0,0 +1,229 @@
...
+
+echo "STARTING HQ SERVER, log in $HQ_SERVER_DIR/HQ.log"
+echo "===================="
+hq server start &>> $HQ_SERVER_DIR/HQ.log &
```
j-kali, line 150: `hq server start &>> "$HQ_SERVER_DIR/HQ.log" &`

same as above, quotation to avoid breaking if path have spaces.

docs/apps/hyperqueue.md
```diff
@@ -0,0 +1,229 @@
...
+echo "STARTING HQ WORKERS ON $SLURM_NNODES nodes"
+echo "===================="
+srun --cpu-bind=none --mpi=none hq worker start --cpus=$SLURM_CPUS_PER_TASK &>> $HQ_SERVER_DIR/HQ.log &
```
j-kali, line 153: `srun --cpu-bind=none --mpi=none hq worker start --cpus=$SLURM_CPUS_PER_TASK &>> "$HQ_SERVER_DIR/HQ.log" &`

same as above, quotation to avoid breaking if path have spaces.

docs/apps/hyperqueue.md
```diff
@@ -0,0 +1,229 @@
...
+
+num_up=$(hq worker list 2>/dev/null | grep RUNNING | wc -l)
```
j-kali, line 155: `num_up=$(hq worker list 2>/dev/null | grep -c RUNNING)`

`grep -c` is enough an cleaner I think

docs/apps/hyperqueue.md
```diff
@@ -0,0 +1,229 @@
...
+while true; do
+    echo "WAITING FOR WORKERS TO START ( $num_up / $SLURM_NNODES )"
+    if [[ $num_up -eq $SLURM_NNODES ]];then
+        break
+    fi
+    sleep 1
+    num_up=$(hq worker list | grep RUNNING | wc -l)
```
j-kali, line 162: `num_up=$(hq worker list 2>/dev/null | grep -c RUNNING)

same as above, `grep -c` is enough an cleaner I think.

But also remove stderr here as wall (since you're doing it already above)

It would be neat to do it at the beginning of the loop (the assignment) to avoid repeating yourself.

docs/apps/hyperqueue.md
```diff
@@ -0,0 +1,229 @@
...
+echo "WAITING FOR JOBS TO FINISH"                                
+sleep 30                                                         
+NOT_DONE=$(hq job list --all | grep "RUNNING\|PENDING")          
+while true; do                                                   
+    echo "WAITING FOR JOBS TO FINISH "                           
+    if [[ -z "$NOT_DONE" ]];then                                 
+        break                                                    
+    fi                                                           
+    sleep 30                                                     
+    NOT_DONE=$(hq job list --all | grep "RUNNING\|PENDING")
```
j-kali, line 179: Somewhat as the other loop, do the assignment at the start of the loop but inside to avoid repeating code.

j-kali (commented):
Somewhat nit-picky comments but example should be hyper correct and clean I think.

rkronberg:
I think this is ok to be published now. Checked that all Jaroslaw's comments had been addressed and polished the style/language a bit. Let's improve this later, e.g. would be nice to have an example also for MPI execution.

docs/apps/hyperqueue.md
```diff
@@ -0,0 +1,241 @@
...
+    resources (task farming) you can just use the CSC utility tool `sbatch-hq`. This way you
+    do not have to care about HyperQueue. Run `module load sbatch-hq` to load the wrapper.
+
+When we have submitted everything we want, we need to wait for the jobs to finish.
+This can be done e.g. with:
+
+```bash
+echo "WAITING FOR JOBS TO FINISH"
+while true; do
+    NOT_DONE=$(hq job list --all | grep "RUNNING\|PENDING")
```
j-kali, line 130: Aside from code repetition, let's just not do that...

How about:

```
while hq job list --all | grep -q "RUNNING\|PENDING" ; do sleep 30 ; done
```

docs/apps/hyperqueue.md
```diff
@@ -0,0 +1,241 @@
...
+the workers early.
+
+Loop until all workers are online (note no timeout):
+
+```bash
+while true; do
+
+    num_up=$(hq worker list | grep -c RUNNING)
+    echo "Checking if workers have started"
+    if [[ $num_up -eq $SLURM_NNODES ]]; then
```
j-kali, line 83: Honestly this `if` should be the loop condition instead of promoting `while true` but it might be hard to read...? 🤔

docs/apps/hyperqueue.md
```diff
@@ -0,0 +1,241 @@
...
+started on compute nodes and execute commands which the client submitted to the server. This
+resembles a Slurm within a Slurm, but you have to start the server and workers yourself.
+
+### Starting the server
+
+!!! info "Note"
+    The following instructions apply for Slurm job scripts where only full nodes
+    are allocated. [A full example](#full-example) (for Mahti) can be found at the bottom.
+
+Specify where the HyperQueue server should be placed. All `hq` commands respect this variable
```
j-kali, line 38: Maybe we could write that the server is holding its state on the filesystem. Not sure if this is clear for a newbie from the text, at least not early enough in the text.

Also, if we are advertising multinode runs as well, we could mention that it needs to be on `lustre`? 🤔 Not sure if this one is necessary, since we probably do not want to suggest even that they could put it on `/dev/shm` for example.

j-kali (commented):
Just few minor comments.

docs/apps/hyperqueue.md
rkronberg: Sure, modified the text a tad to better reflect that the server is on the file system (notably, on lustre).

j-kali (approved):
Looks good to me.

PR #1406, opened 2022-09-22
Just small polishing

sjsillan:
Just small and fast polish before next Monday's workshop.


rkronberg (approved):
Looks good to me. Not a MATLAB expert, but I guess `;` just suppresses output?

PR #1434, opened 2022-10-13
Corrected the example batch job script for Molpro

mariavd:
And added spacing for readability.


rkronberg (approved):
LGTM! For future reference, it would be helpful to include a link to the preview page to expedite the reviewing process :)
In this case, https://csc-guide-preview.rahtiapp.fi/origin/mariavd-molpro-job-script/apps/molpro/

PR #1429, opened 2022-10-07
Created tutorial to install software from Bioconda with Tykky

amsaren:
Created a tutorial to install software from Bioconda with Tykky now that direct conda installations are not available.
https://csc-guide-preview.rahtiapp.fi/origin/saren-bioconda-tykky/support/tutorials/bioconda-tutorial/

PR #1438, opened 2022-10-17
Update tykky.md

JMuff22:
`Tykky is a set of tools for easy installation of tools to HPC using Apptainer containers. ` --> `Tykky is a set of tools which make software installations to HPC easier using Apptainer/Singularity containers.` 

Rewrite such that tools isn't mentioned twice. Consistent use of "Apptainer/Singularity"

Sorry this was missed in the original PR.

### DIFF
--- a/docs/computing/containers/tykky.md
+++ b/docs/computing/containers/tykky.md
@@ -2,7 +2,7 @@
 
 ## Intro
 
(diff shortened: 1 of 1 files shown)


JMuff22:
Resolved in https://github.com/CSCfi/csc-user-guide/pull/1429.
Closed.

PR #1439, opened 2022-10-18
Restricted content width

joonas-somero:
https://csc-guide-preview.rahtiapp.fi/origin/content-width/

Currently, line length on Docs can get awfully long when the width of the browser window grows.

For example, Accounts > Overview:
content_on_the_loose

This PR adresses the issue by restricting content width, like so:
content_tied_down


rkronberg (approved):
Nice!

PR #1440, opened 2022-10-18
Update lue.md

Nortamo:
Specify that date limit only applies to Puhti.


rkronberg:
There was a request that LUE could be more visibly linked to from https://docs.csc.fi/computing/disk/. I'll include that in this branch/PR if ok.

samiilvonen (approved):

PR #1445, opened 2022-10-21
Fixed an issue with loading the environment via the GUI

mariavd:
The AMS GUI does not start an interactive shell when submitting a job, and the environment was not loading properly. It does not recognise the command `module` without sourcing `/appl/profile/zz-csc-env.sh` first.


mariavd:
I noticed that when adding the licence key to the GUI, you get a response from SCM containing your laptop name, mac address and CPU. Should we note the users about it? We don't have to since in principle, a user is supposed to read TOC, but in practice not many do.

rkronberg (approved):

PR #1453, opened 2022-10-28
Restricted content width: a revision

joonas-somero:
https://csc-guide-preview.rahtiapp.fi/origin/content-width-revision/


rkronberg (approved):
Tried to sample pages a bit here and there and looks like nothing is breaking now.

PR #1488, opened 2022-11-23
Increase searchability of weekly support coffees

rkronberg:
https://csc-guide-preview.rahtiapp.fi/origin/coffee/support/contact/

Page can now be found by searching for 'coffee'. 

Also, the announcement bar cannot be dismissed anymore. Is this too intrusive? How the announcement bar works is a bit crude; either it can be dismissed and then it will not return unless the content is changed, or then it cannot be dismissed at all.

### DIFF
--- a/docs/support/contact.md
+++ b/docs/support/contact.md
@@ -6,13 +6,19 @@
 * Email: <mailto:servicedesk@csc.fi>
 * Phone: +358 9 457 2821
 * Open Monday to Friday from 8.30 a.m. to 4 p.m.
-* NEW! Weekly research support session Wednesdays at 2 p.m. More information and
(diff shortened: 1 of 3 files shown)


mlehtiva (approved):
Looks good! I can't say about the announcement bar, except that I personally don't mind it being there.

ktiits:
I would keep the old style, that it is possible "close" the banner. Otherwise looks good.

rkronberg:
Thanks for the comments!

Had a chat with Joonas and we propose to go with not having the option to close the banner. It is rather thin and disappears anyway when you scroll a page. This way we can also ensure that important content stays visible.

Let's revisit this decision later and revert to the old style if needed.

PR #1490, opened 2022-11-24
ssh-keygen use ed25519 instead of rsa

rkronberg:
https://csc-guide-preview.rahtiapp.fi/origin/update-ssh-keys/computing/connecting/

### DIFF
--- a/docs/computing/connecting.md
+++ b/docs/computing/connecting.md
@@ -76,7 +76,7 @@ SSH keys provide more secure authentication and can be enabled with a two-step p
 2. **Copy public key from local PC to supercomputer**  
    Only the _public key_ should be copied, don't copy the private key.
 
-!!! warning "Note"
+!!! error "Important!"
     The private key should **never** be shared with anyone, not even with
     CSC staff. It should be also stored only on the local computer (public key
     can be safely stored in cloud services).
@@ -91,6 +91,13 @@ ssh-keygen -o -a 100 -t ed25519
 
 You will be prompted for a file name and location where to save the
 key. Accept the defaults by pressing `ENTER`.
+
+!!! warning "Note"
+    You will receive a warning if the default file name would overwrite an existing
+    key: `/home/username/.ssh/id_ed25519 already exists.`. Generally, you do not want
+    to overwrite existing keys, so press `n` and provide a different file name manually
+    when prompted. See also the section on [SSH key files with non-default name or
+    location](#ssh-key-file-with-non-default-name-or-location).
 
 Next, you will be asked for a passphrase. Please choose a secure
 passphrase. It should be at least 8 characters long and should contain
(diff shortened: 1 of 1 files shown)


docs/computing/connecting.md
```diff
@@ -1,93 +1,122 @@
...
+below.
 
 ```bash
-ssh-keygen -t rsa -b 4096
+ssh-keygen -o -a 100 -t ed25519
 ```
 
 You will be prompted for a file name and location where to save the
-key. Accept the defaults by pressing ENTER.
+key. Accept the defaults by pressing `ENTER`.
```
CSC-swesters, line 93: We might add a note here that `ssh-keygen` will warn the user if the default file name would overwrite an existing key, in which case the user very like doesn't want to accept that. Then it's better to give a different file name manually.

CSC-swesters: It will look like this, in case you want to make a clear example:
```
$ ssh-keygen -o -a 100 -t ed25519
Generating public/private ed25519 key pair.
Enter file in which to save the key (/home/username/.ssh/id_ed25519): 
/home/username/.ssh/id_ed25519 already exists.
Overwrite (y/n)? n
```

CSC-swesters: Then instead make a more explicit key file name:
```
$ ssh-keygen -o -a 100 -t ed25519
Generating public/private ed25519 key pair.
Enter file in which to save the key (/home/username/.ssh/id_ed25519): /home/username/.ssh/id_ed25519-puhti
Enter passphrase (empty for no passphrase): 
Enter same passphrase again: 
Your identification has been saved in /home/username/.ssh/id_ed25519-puhti.
Your public key has been saved in /home/username/.ssh/id_ed25519-puhti.pub.
The key fingerprint is:
SHA256:7yvhmPa/kdpOh1tWDYw/uGmdyvwuPq9WRg0nFBmnavA username@hostname
The key's randomart image is:
+--[ED25519 256]--+
|             .++.|
|             o+o.|
|          . . += |
|           o +.o.|
|        S   E.+ .|
|        .. + =oo |
|       + .* Boo  |
|      + o= X+.   |
|     . .o+O=*B+  |
+----[SHA256]-----+
```

It would also be useful to link to the section about "SSH key file with non-default name or location" which shows up later.

CSC-swesters (approved):
The changes are great already, thanks for updating this!

I noticed that the docs don't mention the risk of overwriting an existing SSH key pair (by accepting the defaults too much :grin: ), so it would be good to add a comment or even warning about that :)
If you want to do that separately, feel free to merge this PR as is.

docs/computing/connecting.md
rkronberg: Thanks for the suggestion, I'll add this!

CSC-swesters (approved):
Thanks!

PR #1494, opened 2022-11-30
Disable workaround in dockerfile

lvarin:
This hope fixes once and for all the random build errors we have in master


joonas-somero (approved):
Confirmed to work in a development build!

PR #1504, opened 2022-12-13
Clarify automatic cleaning procedures

CSC-swesters:
- Specify that the absence of "path_summary.txt" files is to be expected in some cases, and that a "nothing-to-remove-for-your-project" file will be touched to indicate that this is intentional.
- Add a small bash script which can check all of a user's groups and print their status w.r.t. the automatic cleaning.
- Add some safety features into the bash commands, so that they won't run if the variables are undefined.


CSC-swesters:
Preview: https://csc-guide-preview.rahtiapp.fi/origin/lcleaner-clarifications/support/tutorials/clean-up-data/

joonas-somero (changes requested):

PR #1505, opened 2022-12-16
Update hyperqueue.md

jlento:
https://csc-guide-preview.rahtiapp.fi/origin/jlento-patch-1/apps/hyperqueue/
https://csc-guide-preview.rahtiapp.fi/origin/jlento-patch-1/computing/running/array-jobs/


rkronberg:
Made a separate section for `sbatch-hq` and added an example to clarify usage. I think we should also add a what's new entry to highlight this tool.

Another question concerns the old `sbatch_commandlist` wrapper for array jobs. Is this something we still want to keep around, or does `sbatch-hq` cover all the old use cases of `sbatch_commandlist`?

jlento:
Even better now :)

I'd say sbatch-hq covers the use cases of sbatch_command list. There are probably quite few users of it, so maybe just a box/link that points to sbatch-hq documantation from sbatch_commandlist documentation? Kimmo Mattila probably has the best idea about this. @kkmattil

yetulaxman (approved):

PR #1507, opened 2022-12-19
CCCP-4054: Update instance shelving description

tnykan:
Update instance shelving description that it doesn't free up used resources but instead just frees up those from billing calculations


matteopozza:
https://github.com/CSCfi/csc-user-guide/blame/CCCP-4054/docs/cloud/pouta/vm-lifecycle.md#L64 this is the sentence that has confused the customer, so this one should be fixed. Moreover, here https://github.com/CSCfi/csc-user-guide/blame/CCCP-4054/docs/cloud/pouta/vm-lifecycle.md#L73 it is already mentioned about the billing, so we should avoid repeating the same.

I would suggest something like this:
```
Shelving means shutting down a virtual machine and removing it from the host compute node. This frees up the computing resources that were reserved for the virtual machine. However, the state of all other associated resources, for example the file system, floating IPs, network configuration, etc., is saved in our central storage. Shelving a virtual machine will **not** reduce the amount of resources used by the project, but the billing for that virtual machine will stop.
Shelving works best for the standard flavors that are already backed by our central storage service. Shelving can be slow for flavors that are using local storage, especially bigger flavors since the data needs to be copied between the local and central storage. On a rare occasion, if all of our compute resources are used up, we may not be able to unshelve your virtual machine until another user frees up compute resources. N.B. that your floating IPs, volumes etc. cannot be removed from the virtual machine before it has been unshelved. If you have a floating IP quota of two and one of them is attached to a shelved virtual machine, you have only one left. **Note** that the ephemeral storage in the IO, or TB flavors is **not** shelved.
```

docs/cloud/pouta/vm-lifecycle.md
```diff
@@ -58,25 +58,22 @@ as **active** state virtual machines. Suspending is not generally used
...
-of them is attached to a shelved virtual machine, you have only one
-left. **Note** that the ephemeral storage in the IO, GPU or TB
-flavors is **not** be shelved.
+Shelving means shutting down a virtual machine and removing it from the host compute node. 
+This frees up the computing resources that were reserved for the virtual machine. 
+However, the state of all other associated resources, for example the file system,
+floating IPs, network configuration, etc., is saved in our central storage. 
+Shelving a virtual machine will **not** reduce the amount of resources used by
+the project, but the billing for that virtual machine will stop.
+Shelving works best for the standard flavors that are already backed by our
```
matteopozza, line 67: Can you please add a newline here? It would split the blob of text into two paragraphs, which makes it a bit more readable. It would also highlight a bit more the sentence about the fact that shelving does not reduce the amount of resources used.

matteopozza:
Preview of the page: https://csc-guide-preview.rahtiapp.fi/origin/CCCP-4054/cloud/pouta/vm-lifecycle/

matteopozza (approved):

PR #1513, opened 2022-12-23
Fixed grammar and spelling errors

mariavd:


mariavd:
The text says that by default, `rm` asks for confirmation but it is not the case. Should the alias be added to the global bashrc or should I add a sentence how to make such an alias?

attesillanpaa (approved):

PR #1516, opened 2023-01-02
Updating r-env CmdStan information

helijuottonen:
https://csc-guide-preview.rahtiapp.fi/origin/helijuottonen-cmdstan-update/apps/r-env/#using-r-env-with-stan

Updating the information on within-chain parallelization with brsm and CmdStan: Example on the page refers to using CmdStan backend with brms. Within-chain parallelization (threading) requires a project-specific installation of CmdStan, because it wants to write in the write-protected CmdStan installation folder. Instructions for the CmdStan installation left out, because the installation requires a lot of memory and good enough instructions would take too much space on the page.


jessepharrison (approved):
Nice, agreed that directing people to our helpdesk is probably best for project-specific CmdStan installations.

PR #1517, opened 2023-01-02
Shortened the entry about Materials Studio

mariavd:
Removed the usage and installation instructions.


attesillanpaa:
Maybe also add a whats-new entry in the applications section?

mariavd:
> Maybe also add a whats-new entry in the applications section?

E.g., for new software versions or?

attesillanpaa:
> > Maybe also add a whats-new entry in the applications section?
> 
> E.g., for new software versions or?

That the software is no longer available.

attesillanpaa (approved):

PR #1542, opened 2023-01-18
WIP: EO guide fixing attempt

samumantha:
Trying to fix the EO guide PR revert. 
No new changes to files to review.


ktiits (approved):

samumantha:
Hmm, but this does not include the changes to the other files. Probably because the commits are still in old history.
What does it say is the conflict?
(I cannot see because I do not have write access to master)

PR #1550, opened 2023-01-27
Simplify HQ example

rkronberg:
https://csc-guide-preview.rahtiapp.fi/origin/simplify-hyperqueue/apps/hyperqueue/


jlento:
I think Henrik should review this as the original is from him (we may have also dropped some necessary options). Also, I think both Puhti and Mahti full examples should be available.

rkronberg:
Instead of having two lengthy examples for both Puhti and Mahti, I decided to generalize the existing one. Should be easy to adapt this to either system.

@Nortamo would you have time to check this out?

Nortamo (approved):

PR #1480, opened 2022-11-15
WIP Breadcrumbs and you - getting to know each other

JMuff22:
Trying to fix breadcrumb links in certain page. The broken links are caused by a non-existent url in the breadcrumbs. An easy fix to this is to create an `index.md` file inside every subfolder in `docs/`. 

As an example in `docs/computing/quantum-computing/helmi/first-quantum-job/` the breadcrumbs to `Quantum computing` and `Helmi` are broken. These can be fixed by adding in an `index.md` file in `docs/computing/quantum-computing/`. In this case there in an `overview.md` that should be displayed so the fix can simply be to symbolically link `index.md` to `overview.md` with `ln -s overview.md index.md`. 

The main concern with this solution is that some links to docs.csc.fi will break. 

The page can be previewed here:
https://csc-guide-preview.rahtiapp.fi/origin/fix-breadcrumb-links/

PR #1554, opened 2023-01-30
Add Allas clients to hierarchy

oscarkraemer:


oscarkraemer:
So basically the reasoning for this change is that the pages were not easily findable from left side navigation column. Basically I had never seen these instructions before.

joonas-somero:
It turns out this was exactly what we needed to fix the breadcrumbs navigation in the section in question. This is now merged into #1480 so I'll close this PR.

PR #1573, opened 2023-02-21
WIP Changes to navigation to fix breadcrumbs

joonas-somero:
Retake on a classic!

We noticed that the outcome of #1480 could be achieved with much smaller changes and practically without redirects.

https://csc-guide-preview.rahtiapp.fi/origin/fix-breadcrumbs-navigation/


rkronberg:
A list of some things to be taken care of in following PRs:
- Cloud is missing SectionPage (this could/should be the "Concepts of cloud computing" page), resulting in breadcrumbs missing one level
- Compiling is missing SectionPage (to be added and renamed as Installing, content is already ready)
- Working with data is missing SectionPage (this could/should be the "Contents" page), resulting in breadcrumbs missing one level
- Moving data is missing SectionPage (this could/should be the FAQ entry "How can I move data between CSC and my local computer?"), resulting in breadcrumbs missing one level

rkronberg (approved):

PR #1592, opened 2023-03-03
New version of Turbomole installed

mariavd:
Version update to 7.7 on Puhti and Mahti. Available as the default module.


rkronberg (approved):
ok!

PR #1604, opened 2023-03-09
Remove broken breadcrumb

joonas-somero:
The geoinformatics tutorials happen to be in their own subdirectory. In order for the breadcrumb removed in this PR to work, the directory would need to have an index.md, but since none of the other field-specific tutorials have such an index, the breadcrumb in question should just be removed for consistency.

https://csc-guide-preview.rahtiapp.fi/origin/geoinformatics-tutorials-breadcrumb/support/tutorials/#geoinformatics


rkronberg (approved):
ok

PR #1612, opened 2023-03-16
Remove mkdocs-section-plugin as redundant

joonas-somero:
The feature this plugin provides is already provided by Material for MkDocs and is also enabled by this PR.


rkronberg (changes requested):
Breadcrumbs seems to break by this change. E.g.:
https://csc-guide-preview.rahtiapp.fi/origin/remove-mkdocs-section-plugin/computing/systems-mahti/

joonas-somero:
Whoops! Looks like the feature is implemented differently by the plugin and we need it for the breadcrumbs after all!

PR #1617, opened 2023-03-20
Update kaivos_client_in_puhti.md

kkmattil:
module load mysql converter to module load mairiadb


rkronberg:
I guess this should be merged asap to avoid having incorrect module load commands in docs. I think the `mysql` commands should work as before, since they are symlinked to the corresponding `mariadb` commands. It would still be good that someone who knows about Kaivos/SQL would go through the documentation at some point to verify that everything works as before (with MySQL). @kkmattil @samiilvonen

rkronberg:
Tested a few of the commands and actually not everything seems to be backwards compatible. E.g. could not find `mysql_quota` command and also some of the options of other `mysql` commands do not exist anymore, e.g. `--set-gtid-purged`

@kkmattil would you have time to update the Kaivos documentation?

rkronberg:
Do the Kaivos-pages still need editing or could this PR be reviewed and merged? @kkmattil

kkmattil:
> Do the Kaivos-pages still need editing or could this PR be reviewed and merged? @kkmattil

OK ( we can't test the commands as we don't have a working Kaivos project)

PR #1643, opened 2023-04-21
Better documentation: transfer_data_rahti.md

pitkant:
- Make running the command more understandable, as it was not explicitly mentioned anywhere that POD needs to be replaced with user input (neither is local/dir and remote/dir but they are more established placeholders)
- Add missing tick marks


joonas-somero:
Thank you for your contribution to Docs CSC, @pitkant!

joonas-somero (approved):

PR #1650, opened 2023-04-27
Fix typo in multi-stage-builds.md

pitkant:
FORM -> FROM
dockerfile -> Dockerfile


joonas-somero (approved):
Thanks again, @pitkant!

PR #1653, opened 2023-05-02
docker push needs sudo?

samumantha:
I am not sure if this worked earlier without sudo but now it seems to need it?


yetulaxman:
> I am not sure if this worked earlier without sudo but now it seems to need it?

Yes. it needs sudo. I have at least been using with sudo

yetulaxman (approved):

PR #1657, opened 2023-05-05
Added comment about objects per bucket quota

samiilvonen:


samiilvonen:
Got approval outside github

PR #1659, opened 2023-05-10
Single register update

lkalliok:
Updated the guidelines for registers and added instructions for data access with SD Apply.


kkmattil (approved):
Looks good!

PR #1671, opened 2023-05-19
Separate sensitive data examples into a bullet list

aajarven:
The paragraph with an inline "bullet list" was somewhat difficult to read and grammatically problematic. I separated the items to a proper bullet list for easier reading.

I also softened the wording in the text leading to the bullet list so that it does not sound like everything in the list is always sensitive (e.g. "There are narcissus flowers at Esplanadi park" is ecological information, but hardly sensitive) and that nothing outside the list is.


fmorelloCSC:
Hi. That definition is by GDPR the definition of sensitive data and all those data type are subject to a very strict legislation to which both users and us (as service provide) need to comply. 
I am not sure if I mentioned before, but we currently have 3 interns that are working on improving docs (both text and structure) and they are planning to re-write re-organize the introduction completely, so any additional suggestion is very welcome. Thanks for your help!

fmorelloCSC (approved):

PR #1681, opened 2023-05-26
Added R 4.2.2

helijuottonen:
Added R 4.2.2 to r-env documentation and a mention of the update to the front page. 
In addition in r-env documentation: 
1) Added a section for solving the markdown pdf problem.
2) Added a mention to 'R package installations' that modifying R_LIBS only works when not using RStudio (because in 4.2.2 .Renviron is in tmpdir, not in home directory).
3) Updated R version and CmdStan version in the CmdStan instructions.

### DIFF
--- a/docs/apps/r-env.md
+++ b/docs/apps/r-env.md
@@ -751,7 +751,7 @@ To use R packages installed in `/projappl`, add the following to the beginning o
 .libPaths(c("/projappl/<project>/project_rpackages_<rversion>", .libPaths()))
 ```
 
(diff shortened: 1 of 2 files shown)


jessepharrison (commented):
One small suggestion before merging: 
L754: `... or when using the R console)` --> `... or when using the command-line version of R` (or some such...).

RStudio also has a console, so this could help clarify that we are referring to non-RStudio work.

jessepharrison (approved):

PR #1699, opened 2023-06-13
Further development on command line tutorial

joonas-somero:
https://github.com/CSCfi/csc-user-guide/blob/update-windows-tutorial/GETTING_STARTED.md (renamed from WINDOWS.md)


rkronberg:
Looks good, but I wonder whether it would make sense to keep the stuff that is specific to windows in this page and move the OS-agnostic git commands to some other tutorial (or extend/improve the general contributing guide)?

joonas-somero:
> Looks good, but I wonder whether it would make sense to keep the stuff that is specific to windows in this page and move the OS-agnostic git commands to some other tutorial (or extend/improve the general contributing guide)?

When I was in a hurry to have the Git part available _somewhere_, I felt that this file is as good as any, but now that you brought it up, maybe I'll try to moving it somewhere else and see how it fits there. Or, I've named this file WINDOWS.md, but that's misleading in a sense, since the tutorial is not so much about Windows itself than it is about getting absolute beginners started with the command line. It's just that the intended audience have likely chosen Windows as their OS. Or that's what I assume, at least. So simply renaming the file (and perhaps explicitly pointing out the OS-agnostic parts) would be another solution.

rkronberg:
Yes! I think the latter is a good solution :+1:

joonas-somero:
> Yes! I think the latter is a good solution +1

I gave it a try! Feel free to improve on anything!

mariavd:
I tested the whole tutorial on my Windows computer where I didn't previously have Conda, KeePass and Git bash installed. It worked smoothly for the most part. 

I added a paragraph, suggesting a modification to the Hotkeys section. I believe the previous one was too complicated.

mariavd:
Some further comments:

_A security warning dialog will most likely pop up asking for confirmation on running the file._

Maybe “may appear” because I have disabled these warnings on my computer and I don't get any of them. 

On Windows 10, my default Git installation path is C:\Program Files\Git

_and sooner or later Windows will find it for you._

It happens right away. 

_clicking "Yes" if Notepad asks to create the file. Then, copy-paste the following lines into the file and save it_

The first part of the script is already given in the GitHub link

And also, my Conda script is in C:\Users\<username>\miniconda3\etc\profile.d

joonas-somero:
> Maybe “may appear” because I have disabled these warnings on my computer and I don't get any of them.

I was making some vague assumption about default settings.

> On Windows 10, my default Git installation path is C:\Program Files\Git

'C:\Program Files' shouldn't be accessible as a non-admin user.

> It happens right away.

It might not. I've experienced Windows not suggesting something even after I've typed an exact match. :slightly_smiling_face:

Or, does "sooner or later Windows will find it for you" give the impression that there is some kind of latency with the search? What I meant was that when inputting the name of the program one letter at a time, Windows will "sooner or later" (well, hopefully) suggest the correct one. On the system I'm working on this, typing `g` suggest Snipping Tool, Google Chrome and so on, but Git Bash is nowhere to be seen. After `gi`, the top suggestion is Git Bash.

> The first part of the script is already given in the GitHub link

Yes, right. I suppose I meant to only mention that the script is copied from that tutorial. That'll need to be reworded or something.

> And also, my Conda script is in C:\Users<username>\miniconda3\etc\profile.d

Perhaps this needs to be checked before adding to the script. Something like "if output of `head ~/.../conda.sh` has _No such file or directory_, then `conda.sh` is elsewhere.

mariavd (approved):
Looks good to me.

PR #1710, opened 2023-06-28
Add object storage to allas instructions for rclone

samumantha:
Copying files from another object storage to Allas has been requested once in a while, and will probably increase in future especially for users of earth observation data. 

"Researcher me" got these instructions some time ago from @kkmattil and I thought it might be nice to have them public in docs pages, what do you think?

This may need some more information on how to get the data to be filled in the config file?
-> is that always something like `openstack ec2 credentials list` ?


kkmattil (approved):
OK

PR #1731, opened 2023-07-18
Hyperqueue improvements

jaantollander:
Improvements to the HyperQueue documentation.

- Major refactoring and structural improvements, the explanation is now structured around the example
- The example is generalized for partial and full node allocations
- Added a section about using local disk with HyperQueue

Unfortunately, we require some control flow in Bash to set the memory resources for HyperQueue.
Hopefully there is a way to get rid of it in the future. 

Preview: https://csc-guide-preview.rahtiapp.fi/origin/hyperqueue/apps/hyperqueue/


docs/apps/hyperqueue.md
```diff
@@ -4,234 +4,280 @@ tags:
...
+=== "Puhti partial multinode"
+    ```bash
+    #!/bin/bash
+    #SBATCH --account=<project>
+    #SBATCH --partition=large    # multi node partition
+    #SBATCH --ntasks=2           # two or more HyperQueue workers
+    #SBATCH --cpus-per-task=10   # one or more cpus per worker
+    #SBATCH --mem-per-cpu=1000   # desired amount of memory
+    #SBATCH --time=00:15:00
+    ```
```
rkronberg, line 110: Perhaps
```bash
#SBATCH --nodes=2
#SBATCH --ntasks-per-node=1
```
would be more clear?

docs/apps/hyperqueue.md
```diff
@@ -4,234 +4,280 @@ tags:
...
+
+=== "Puhti full multinode"
+    ```bash
+    #!/bin/bash
+    #SBATCH --account=<project>
+    #SBATCH --partition=large
+    #SBATCH --ntasks=2           # two or more worker nodes
+    #SBATCH --cpus-per-task=40   # reserve all cpus on a node
+    #SBATCH --mem-per-cpu=0      # reserve all memory on a node
+    #SBATCH --time=00:15:00
```
rkronberg, line 131: See above

docs/apps/hyperqueue.md
```diff
@@ -4,234 +4,280 @@ tags:
...
+until hq job list &> /dev/null ; do sleep 1 ; done
 ```
 
-This is a non-blocking command similar to `sbatch`.
+**Workers**
 
-HyperQueue is not limited to running a single execution per submission. Using the
-`--array 1-N` flag we can start a program *N* times similar to how Slurm array jobs work.
+Next, we start HyperQueue workers in the background with the number of CPUs and the amount of memory defined in the batch script.
+We access those values using the `SLURM_CPU_PER_TASK` and `SLURM_MEM_PER_CPU` environment variables.
```
rkronberg, line 182: Typo, should be `SLURM_CPUS_PER_TASK`

docs/apps/hyperqueue.md
```diff
@@ -4,234 +4,280 @@ tags:
...
-hq submit --array 1-10 --cpus <n> <COMMAND>
+# Set memory for workers in bytes according to SLURM_MEM_PER_CPU if greater than zero.
+# Otherwise, leave unset to use all the memory of the node.
+if [[ "${SLURM_MEM_PER_CPU:-0}" -gt 0 ]]; then
+    # Calculate the total memory reservation and convert it from megabytes to bytes.
+    TOTAL_MEM_BYTES=$((SLURM_CPUS_PER_TASK * SLURM_MEM_PER_CPU * 1000000))
+    TOTAL_MEM_OPT="--resource mem=sum($TOTAL_MEM_BYTES)"
+else
+    TOTAL_MEM_OPT=""
+fi
```
rkronberg, line 195: Is this really necessary? I've used hyperqueue without this and haven't noticed memory issues.

docs/apps/hyperqueue.md
```diff
@@ -4,234 +4,280 @@ tags:
...
+
+# Start the workers in the background.
+srun --overlap --cpu-bind=none --mpi=none hq worker start \
```
rkronberg, line 198: Previous example used `--exact`. Any specific reason for using `--overlap here`?

docs/apps/hyperqueue.md
```diff
@@ -4,234 +4,280 @@ tags:
...
 
-#### sbatch-hq
 
-For simple task farming workflows, where you only want to run many similar (non-MPI parallel,
-independent) commands/programs, you can use the CSC utility tool `sbatch-hq`. Just specify the
-list of commands to run in a file, one command per line. The tool `sbatch-hq` will create and
-launch a batch job that starts running commands from the file using HyperQueue. You can specify
-how many nodes you want to run the commands on and `sbatch-hq` will keep executing the commands
-until all are done, or the batch job time limit is reached.
+### Using local disks with HyperQueue
```
rkronberg, line 238: I wonder if this should be extended with a more complete batch script example?

rkronberg (changes requested):
Nice work! Few comments below. I also wonder if there should still be one complete copy-pasteable batch script example as in the old version?

docs/apps/hyperqueue.md
rkronberg: More generally, why reserve 2 nodes with 10 threads instead of 1 node with 20 threads?

docs/apps/hyperqueue.md
jaantollander: This code is indeed quite ugly.
If HyperQueue works well without this, we should remove it.
I suppose it is up to the programmer to not exceed the memory limit.

docs/apps/hyperqueue.md
jaantollander: I was thinking that it is easier for Slurm to allocate a smaller fraction of CPUs on a node to more tasks (processes) so that slurm can decide where to place them.
Since HyperQueue tasks don't communicate with one another, it shouldn't matter if they run of workes on different nodes.

docs/apps/hyperqueue.md
```diff
@@ -4,234 +4,280 @@ tags:
...
+If a server directory is not specified, it will default to the user's home directory.
+In this case, one has to be careful not to mix up separate computations. We recommend starting one server per job in a job-specific directory for simple cases that fit inside one Slurm job.
 
 ```bash
-hq worker wait "${SLURM_NTASKS}"
-```
-
-### Submitting jobs
+# Specify a location for the server
+export HQ_SERVER_DIR="$PWD/.hq-server/$SLURM_JOB_ID"
```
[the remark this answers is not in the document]
jaantollander: I can remove it.

docs/apps/hyperqueue.md
jaantollander: I can double check this.
HyperQueues autoallocator used `--overlap` so I used it as well.

docs/apps/hyperqueue.md
jaantollander: I can include a more extensive example. I'll add it to the full copy-pasteable scripts.

jaantollander:
I made implemented some of the fixes and added example scripts.

rkronberg:
Thanks! I'll review today or latest on Monday.

docs/apps/hyperqueue.md
rkronberg: I'd propose to remove this for now in order not to make the example to complicated.

rkronberg (approved):

PR #1740, opened 2023-07-27
Update how-to-create-new-user-account.md

jotaskin:
There was some old info about student accounts:
There are three different types of CSC user accounts: ordinary accounts for general use (which you can get with or without Haka or Virtu), student accounts for courses, and machine-to-machine robot accounts for managing services.


jarnolaitinen (approved):
I don't know about the student accounts, but I trust Johanna knows :)

PR #1748, opened 2023-08-01
Buttons for simple feedback on specific page

joonas-somero:
https://csc-guide-preview.rahtiapp.fi/origin/feedback-buttons/

image

After clicking on "Yes":
image

After clicking on "No":
image

Prefilled title and label in new issue:
Feedback


rkronberg:
Text suggestions:
* **Yes**:  
Thank you for you feedback!
* **No**:  
Thank you for your feedback! Please let us know how we could improve this page by [contacting us](https://docs.csc.fi/support/contact/) or by [creating an issue in GitHub](https://github.com/CSCfi/csc-user-guide/issues/new). You can also [edit this page yourself](https://github.com/CSCfi/csc-user-guide/edit/master/docs/index.md) and contribute your improvements through a pull request.

The link to editing would of course be different for each page (same as when clicking the pen icon at the top of a page).

rkronberg:
Maybe also the 'Do you find this page useful?' could be written in past tense 'Did you find this page useful?' I guess you need to first read the page in order to then decide whether it was useful or not?

joonas-somero:
> Maybe also the 'Do you find this page useful?' could be written in past tense 'Did you find this page useful?' I guess you need to first read the page in order to then decide whether it was useful or not?

I'm viewing the site as a handbook where pages are constantly revisited, so the idea here is that the page presently exists and is either useful or not.

rkronberg:
> > Maybe also the 'Do you find this page useful?' could be written in past tense 'Did you find this page useful?' I guess you need to first read the page in order to then decide whether it was useful or not?
> 
> I'm viewing the site as a handbook where pages are constantly revisited, so the idea here is that the page presently exists and is either useful or not.

Fair enough, that makes also sense!

joonas-somero:
> > > Maybe also the 'Do you find this page useful?' could be written in past tense 'Did you find this page useful?' I guess you need to first read the page in order to then decide whether it was useful or not?
> > 
> > 
> > I'm viewing the site as a handbook where pages are constantly revisited, so the idea here is that the page presently exists and is either useful or not.
> 
> Fair enough, that makes also sense!

Oh, and when the page source file changes, the previous selection is removed and the buttons shown again. That makes it an on-going dialog, rather than a one time "yes" or "no".

rkronberg:
Preview of this branch seems to not be building correctly. Other branches are fine, so I wonder if this is an issue with the changes made here and whether those could also be carried over to production?

joonas-somero:
> Preview of this branch seems to not be building correctly. Other branches are fine, so I wonder if this is an issue with the changes made here and whether those could also be carried over to production?

Looks fine now. :man_shrugging: Perhaps preview fixed itself over the weekend.

jemaltahir (commented):
Look at the minor suggestion, if you consider there are benefits you can change it. Otherwise it's LGTM (JS) files.

jemaltahir (commented):
Changes have been applied.

PR #1749, opened 2023-08-03
Add note on exeeding available memory/time for partition

samumantha:
Based on brief discussion in RC.

Not sure of the latter part of the new sentence is needed or maybe could be formulated nicer?


rkronberg (approved):
Looks good, thanks! Reworded the last part a bit.

PR #1774, opened 2023-08-29
Add script for checking possibly outdated pages

rkronberg:
Suggestion: Add script for checking possibly outdated files. Running `bash scripts/last_update.sh` prints a sorted list of files that have been last edited by the user. Add instructions into `CONTRIBUTING.md` to run this every now and then and check files that have not been touched in a long time after the user's latest commit.

Does this make sense? It could also be made so that the script prints all pages (not just those that have been last edited by the user). Filtering by user could be done manually, e.g. `bash scripts/last_update.sh | grep "$(git config user.name)"`


joonas-somero:
Should the script look into commit messages as well to find commits co-authored by the user? Perhaps only optionally with a switch.

> Does this make sense? It could also be made so that the script prints all pages (not just those that have been last edited by the user). Filtering by user could be done manually, e.g. `bash scripts/last_update.sh | grep "$(git config user.name)"`

How about passing user.name as an argument?

rkronberg:
> Should the script look into commit messages as well to find commits co-authored by the user? Perhaps only optionally with a switch.
> 
> > Does this make sense? It could also be made so that the script prints all pages (not just those that have been last edited by the user). Filtering by user could be done manually, e.g. `bash scripts/last_update.sh | grep "$(git config user.name)"`
> 
> How about passing user.name as an argument?

Added command-line option for the username. Not sure whether co-authors are needed, if you figure out how to include those easily go ahead, but for now I think personally this solution would be enough. We can continue developing this as needed later also.

joonas-somero (approved):

PR #1779, opened 2023-09-01
Improvements to the Julia documentation

jaantollander:
- Removed unnecessary `srun`s. We use Julia to manage resources and launch new processes instead of `srun`.  Only MPI execution requires `srun`, which is now called from a more portable Julia wrapper.
- Updated Julia versions on the project dependencies.
- Removed `job-name` from batch scripts.
- Improved language regarding MPI such that it takes LUMI into account.


sjsillan (approved):
Hyvältä näyttää!

PR #1791, opened 2023-09-08
Fm retentionupdate

fmorelloCSC:
Updated account section (retention for SD Desktop in billing units)
Updated what's new
Updated SD Desktop standard and Findata with Ubuntu OS


kkmattil (approved):
OK

PR #1802, opened 2023-09-19
SSH key sync delay fix

rkronberg:
https://csc-guide-preview.rahtiapp.fi/origin/ssh-patch/computing/connecting/#adding-a-public-key-in-mycsc-portal


CSC-swesters (approved):
Looks good, thanks!

PR #1809, opened 2023-09-28
CSC Notebooks: CSC login autopromotion

tourunen:
Change teacher guide to reflect the fact that we automatically assign workspace quota to any new users logging in using their CSC accounts.


katrite (approved):
Looks good!

PR #1815, opened 2023-10-05
The handling of the authentication token for helmi is changing, and i…

tiggi:
…t will require the jobs to run with a login shell. Add -l to the example ahead of time

Adding -l forces the job to read ~/.bashrc, and that is required to find the project-specific authentication token.


rkronberg:
Close, not relevant anymore.

PR #1831, opened 2023-10-16
Update mkdocs.yml

ainoc:
Moved Sensitive data to main level of side navigation.
Preview: https://csc-guide-preview.rahtiapp.fi/origin/ac-sdupdate/


fmorelloCSC:
We changed the navigation bar in a new fresh branch, moved sensitive data to main menu to guide users in a better way, based on use cases.

The first page needs links fixed (in the boxed, the are all broken): docs/data/sensitive-data/index.md

 Can docs people please help with it?

fmorelloCSC:
Thank you! It looks amazing. Could you please remove "Tutorials and quick guides". We need to think about that a little bit more.

fmorelloCSC:
We added new files (because using the same link was twice in the menu and it was making the menu jump to the wrong  section), but now the preview does not work anymore. Travis does not see any problem, Rahti does not create the preview. 
image

This was the last fix, it can then go out.

kkmattil (commented):
In command line encryption example private key is used for signing encryption. Is that required as it may confuse some users?

crypt4gh encrypt --sk mykey.sec

fmorelloCSC:
We did not change the text in that section, so if you want to update it and take off the private key for me it is ok.

kkmattil (approved):

rkronberg (approved):
Technical side also ok now!

PR #1836, opened 2023-10-19
typo fix

jpcsc:


matteopozza:
Fixing a long lasting bug in our documentation, i.e. a space in the name of a vm flavor with a space between the number of cores and the word "cores" itself. Convention is that flavor names do not have spaces.

matteopozza (approved):

PR #1843, opened 2023-10-24
Update using_ida.md

spousi:
Removed IDA use based on Academy of Finland funding decision (no longer an option). Added a sentence about Qvain in the beginning.


kkmattil (approved):
OK

PR #1853, opened 2023-10-30
Some documentation and cleaning up

joonas-somero:
https://csc-guide-preview.rahtiapp.fi/origin/minor-technical-changes-2023-10-30/

Get some housekeeping out of the way in preparation for implementing new features:
- Consistency for code formatting (excluding JavaScript, that's for another PR)
  - `'` -> `"`
  - indentation
  - import rest of stylesheets in _extra.css_ instead of having a long list of them in _mkdocs.yml_
    - this is also consistent with upcoming reorganization of JavaScript into modules
- Facilitate the use of snippets
  - move the glossary into it's own directory
  - documentation
- Adjustments in Reference card
  - try to get it closer to a model document for best practices


rkronberg (changes requested):
Nice! However, this seems to have broken the tooltips for abbreviations in the glossary. Not sure why..

joonas-somero:
> Nice! However, this seems to have broken the tooltips for abbreviations in the glossary. Not sure why..

WOMM! :smile: This may have something to do with Preview. One thing I at least noticed is that the script for generating the glossary page was never added into the list of scripts to run, so the page isn't being generated. That shouldn't affect the abbreviations, though.

joonas-somero:
The script itself had the old path, but that made no difference regarding the tooltips. I tried running the branch in a container and they're not working (before or after fixing the script), so it' s not Preview.

joonas-somero:
The problem was 688e770, but now _Travis_ can't find the package with that version.

joonas-somero:
With updates to Travis configuration, dependencies and [Preview](https://github.com/CSCfi/docs-preview/pull/29), everything seems to be working.

docs/ref.md
```diff
@@ -64,10 +70,12 @@ There is currently no special mechanism in place for controlling banners.
     -->
     ```
 
+
 ## Glossary
+
 There is a glossary of HPC-related acronyms that get highlighted automatically. For example: CPU,
 GPU, QPU, etc. The acronyms are defined in the markdown file
-`csc-overrides/assets/glossaries/hpc.md`. More acronyms (case-sensitive) can be added there or into another
+_csc-overrides/assets/snippets/glossaries/hpc.md_. More acronyms (case-sensitive) can be added there or into another
```
rkronberg, line 78: What's our current policy on using inline code blocks? Shouldn't paths be formatted as code?

joonas-somero: Perhaps this should have been discussed. :thinking: Anyway, my opinion is to leave the inline and fenced code specifically for examples of actual code or commands. When referencing a name, even when the name usually appears in some code (and even when it happens to be prefixed with a path), I would use something else to format it. For example:

The static method _Classname.static_method_ is used like this:
```python
Classname.static_method(arg)
```

The file _~/scripts/script.sh_ needs to have the execution permission set:
```bash
chmod u+x ~/scripts/script.sh
```

I'm actually not sure whether boldface would be even better. In any case, looks like I'm not even abiding by my own rules in the Snippets section, where I've used (or probably forgot to reformat) inline code.

rkronberg (approved):

PR #1904, opened 2023-11-27
Newsletter/mailing list archives

joonas-somero:
[Archives listed on 'Contact' page](https://csc-guide-preview.rahtiapp.fi/origin/newsletter-archive/support/contact/#archives)


joonas-somero:
The FIXME (should be the only thing failing the tests) will need to be removed before merging to master.

mkdocs.yml
```diff
@@ -5,6 +5,7 @@ site_url: "https://docs.csc.fi/"
 site_description: Instructions and user guides for the CSC supercomputers, cloud services, storage and software, including FAQ and tutorials
 
 extra:
+  environment: !ENV [MKDOCS_ENV, "production"] # FIXME to this: environment: !ENV [MKDOCS_ENV, "preview"]
```
rkronberg, line 8: Comment still relevant?

rkronberg (approved):
One small question about a comment. Otherwise great work! Btw, why was the FIXME required to force tests to fail?

mkdocs.yml
joonas-somero: The comment is what I'll replace the key with momentarily.

joonas-somero:
> One small question about a comment. Otherwise great work! Btw, why was the FIXME required to force tests to fail?

I wanted to not have the tests pass until the workaround to show the archives in Preview (for the review) is removed, so I added a FIXME. Well, actually the tests aren't looking for FIXMEs in _mkdocs.yml_, so I added another one in _support/contact.md_. Anyway, I'll remove them now, and with that, the archives are only fetched when an environment variable `MKDOCS_ENV=production` is defined.

PR #2031, opened 2024-03-14
Update vasp.md

jlento:
Added Jupyter Lab instructions for py4vasp

https://csc-guide-preview.rahtiapp.fi/origin/jlento-patch-1/


rkronberg (approved):
LGTM! But would it make sense to update the batch script example also to use Mahti, as it is now stated that the page describes how to use VASP on Mahti, not Puhti.

jlento:
Oh, forgot Puhti in the section header :) Gone now.

PR #2041, opened 2024-03-25
Speed up build

oscarkraemer:
## Proposed changes

- This changes tires to speed up the build process.
- It assumes that  requirements.txt will seldom be changed and therefor it can use a cached version.
- Removing the git clone, as far as I can see there is no use of this clone BUT I might be wrong.
- There should be no changes to https://csc-guide-preview.rahtiapp.fi/origin/testing-speedup/
- It assumes that the /docs will be changed most frequently and therefor everything that is dependant on it is removed to the end of the Dockerfile

- [x ] I have followed the instructions in the [Contributing](https://github.com/CSCfi/csc-user-guide/blob/master/CONTRIBUTING.md) and [Styleguide](https://github.com/CSCfi/csc-user-guide/blob/master/STYLEGUIDE.md) documents.
- [ x] My pull request passes all tests.
- [ x] I have included a link to the appropriate [preview page](https://csc-guide-preview.rahtiapp.fi/origin/) (select your branch from the list).

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -14,30 +14,22 @@ RUN yum -y install epel-release &&\
     yum -y install findutils &&\
     yum clean all
 
-RUN chgrp -R root ${ROOT_GROUP_DIRS} &&\
-    chmod -R g+rwx ${ROOT_GROUP_DIRS}
+
+COPY requirements.txt /requirements.txt
+RUN chgrp -R root ${ROOT_GROUP_DIRS} && \
+    chmod -R g+rwx ${ROOT_GROUP_DIRS} && \
+    pip3 install --no-cache-dir -r /requirements.txt
+
+COPY nginx.conf /etc/nginx
+EXPOSE 8000
+CMD [ "/usr/sbin/nginx" ]
 
 COPY . /tmp
-
 WORKDIR /tmp
 
-
-RUN git clone --no-checkout https://github.com/CSCfi/csc-user-guide git_folder && \
-    if [ -d ".git" ]; then rm -r .git; fi && \
-    mv git_folder/.git . && \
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -14,30 +14,22 @@ RUN yum -y install epel-release &&\
...
 
+COPY nginx.conf /etc/nginx
+EXPOSE 8000
+CMD [ "/usr/sbin/nginx" ]
 
-RUN git clone --no-checkout https://github.com/CSCfi/csc-user-guide git_folder && \
-    if [ -d ".git" ]; then rm -r .git; fi && \
-    mv git_folder/.git . && \
-    rm -r git_folder && \
-    git reset HEAD --hard && \
```
lvarin, line 29: AFAIK this is still needed to generate the modification dates in the pages.

Screenshot from 2024-03-25 14-27-51

lvarin: Which is not present in the build done by Rahti BC. Keep in mind that the preview and the production are very different processes.

oscarkraemer: 🤔 the git log says that the page was modified March 7 2023. 

The Last update  seems to be the same on the feature branch as well as master:
https://csc-guide-preview.rahtiapp.fi/origin/testing-speedup/cloud/
https://docs.csc.fi/cloud/

There is of course something that I still might be missing.

I'm also wondering that IF it is still needed maybe it would be possible to use `ADD --keep-git-dir=true` instead of `COPY . /tmp`

lvarin: In a Build config OpenShift clones the code automatically but removes the `.git` directory before starting to run `Dockerfile`. I can prepare a demo example later today/tomorrow.

PR #2081, opened 2024-05-07
New tutorial for rahti proxy

lvarin:
## Proposed changes

Briefly describe the changes you've made here. Remember to add a link to the [preview page](https://csc-guide-preview.rahtiapp.fi/origin/) of your branch.

- [ ] I have followed the instructions in the [Contributing](https://github.com/CSCfi/csc-user-guide/blob/master/CONTRIBUTING.md) and [Styleguide](https://github.com/CSCfi/csc-user-guide/blob/master/STYLEGUIDE.md) documents.
- [ ] My pull request passes all tests.
- [ ] I have included a link to the appropriate [preview page](https://csc-guide-preview.rahtiapp.fi/origin/) (select your branch from the list).


docs/cloud/tutorials/integrating-external-services.md
```diff
@@ -0,0 +1,114 @@
+# Integrating External Services
+
+Kubernetes, and by extension OpenShift OKd, gives a lot of flexibility regarding network use cases. One of the uses cases is the one that allows to use external services, like databases, transparently in a OpenShift project. Other of the use cases would be to to have a network proxy between OpenShift installations. This could be used as a temporal to ease migration periods.
+
```
trispera, line 4: Missing word?
`This could be used as a temporal to ease migration periods`
>
`This could be used as a temporal solution to ease migration periods`

docs/cloud/tutorials/integrating-external-services.md
```diff
@@ -0,0 +1,114 @@
...
+    spec:
+      host: <test>.rahtiapp.fi
+      port:
+        targetPort: http
+      to:
+        kind: Service
+        name: <Service-name>
+        weight: 100
+      wildcardPolicy: None
+    status: {}' | oc create -f -'
```
trispera, line 31: This command worked for me:
```yaml
echo "apiVersion: route.openshift.io/v1
kind: Route
metadata:
  name: test-route
spec:
  host: <test>.rahtiapp.fi
  port:
    targetPort: http
  to:
    kind: Service
    name: <Service-name>
    weight: 100
  wildcardPolicy: None" | oc create -f -
```

docs/cloud/tutorials/integrating-external-services.md
```diff
@@ -0,0 +1,114 @@
...
+          targetPort: 80 
+          nodePort: 0
+    selector: {}' | oc create -f -
+    ```
+    The name of the `Service` and the `EndPoint` must be the same.
+
+1. Finally the `Route` in Rahti 1:
+
+    ```yaml
+    echo 'apiVersion: route.openshift.io/v1
```
trispera, line 87: I tried with our `course-flask-demo`. I don't know if it's the case for all the applications (maybe most of them), should we add:
```yaml
spec:
  ...
  tls:
    insecureEdgeTermination: Redirect
    termination: edge
  ...
```
?

docs/cloud/tutorials/integrating-external-services.md
lvarin: yes

docs/cloud/tutorials/integrating-external-services.md
```diff
@@ -0,0 +1,114 @@
+# Integrating External Services
+
+Kubernetes, and by extension OpenShift OKd, gives a lot of flexibility regarding network use cases. One of the uses cases is the one that allows to use external services, like databases, transparently in a OpenShift project. Other of the use cases would be to to have a network proxy between OpenShift installations. This could be used as a temporal to ease migration periods.
```
lvarin:
```suggestion
Kubernetes, and by extension OpenShift OKd, gives a lot of flexibility regarding network use cases. One of the uses cases is the one that allows to use external services, like databases, transparently in a OpenShift project. Other of the use cases would be to to have a network proxy between OpenShift installations. This could be used as a temporal solution to ease migration periods.
```

docs/cloud/tutorials/integrating-external-services.md
lvarin: True, Rahti 1 has the certificate, but Rahti 2 does not.

lvarin: Does this make more sense?

    Replace `<test>` so the URL is the same as in Rahti 1. Rahti 1 will be later configured to relay the request to Rahti 2, when the request reaches Rahti 2, the Loadbalancer needs to know to which application to send the request, the host parameter is used (Using `HTTPD` headers).

lvarin:
All changes applied, and an extra one to check

docs/cloud/tutorials/integrating-external-services.md
trispera: I like this change. Still a difficult concept to explain in details. This one is a good approach 👍

trispera (approved):
LGTM

PR #2151, opened 2024-07-10
apps update

xaprasanna:
Reviewed /docs/apps pages.


docs/apps/wtdbg2.md
```diff
@@ -23,13 +23,13 @@ Version on CSC's Servers
...
 
-In Puhti, the wtdbg2 is activated by loading the biokit environment.
+On Puhti, wtdbg2 is activated by loading the biokit environment.
 
 ```text
 module load biokit
 ```
 
-After that you can use commands `wtdbg2` asembler and `wrpoa-cns` consenser commands. _wtdbg2_ assembles raw reads and generates the contig layout and edge sequences in a file _prefix_.ctg.lay.gz. Executable _wtpoa-cns_ takes this file as input and produces the final consensus in FASTA. 
+After that, you can use commands `wtdbg2` asembler and `wrpoa-cns` consenser commands. _wtdbg2_ assembles raw reads and generates the contig layout and edge sequences in a file _prefix_.ctg.lay.gz. Executable _wtpoa-cns_ takes this file as input and produces the final consensus in FASTA.
```
amsaren, line 32:
```suggestion
After that, you can use commands `wtdbg2` asembler and `wrpoa-cns` consenser commands. `wtdbg2` assembles raw reads and generates the contig layout and edge sequences in a file _prefix_.ctg.lay.gz. Executable `wtpoa-cns` takes this file as input and produces the final consensus in FASTA. 
```

docs/apps/wtdbg2.md
```diff
@@ -39,11 +39,11 @@ wtdbg2 -x rs -g 4.6m -t 16 -i reads.fa.gz -fo prefix
 wtpoa-cns -t 16 -i prefix.ctg.lay.gz -fo prefix.ctg.fa
 ```
 
-In the wtdbg2 command `-g` is the estimated genome size and `-x` specifies the sequencing technology, which could take value _rs_ for PacBio RSII, _sq_ for PacBio Sequel, _ccs_ for PacBio CCS reads and _ont_ for Oxford Nanopore. This option sets multiple parameters and should be applied before other parameters. When you are unable to get a good assembly, you may need to tune other parameters as described in the wtdbg2 manual.
+In the `wtdbg2` command, `-g` is the estimated genome size and `-x` specifies the sequencing technology, which could take value _rs_ for PacBio RSII, _sq_ for PacBio Sequel, _ccs_ for PacBio CCS reads and _ont_ for Oxford Nanopore. This option sets multiple parameters and should be applied before other parameters. When you are unable to get a good assembly, you may need to tune other parameters as described in the wtdbg2 manual.
```
amsaren, line 42:
```suggestion
In the `wtdbg2` command, `-g` is the estimated genome size and `-x` specifies the sequencing technology, which could take value `rs` for PacBio RSII, `sq` for PacBio Sequel, `ccs` for PacBio CCS reads and `ont` for Oxford Nanopore. This option sets multiple parameters and should be applied before other parameters. When you are unable to get a good assembly, you may need to tune other parameters as described in the wtdbg2 manual.
```

amsaren (changes requested):

kkmattil (approved):
OK

docs/apps/bioperl.md
```diff
@@ -31,15 +31,15 @@ BioPerl is licensed under the same terms as Perl itself which is dually-licensed
 
 ## Available
 
--   Puhti: Perl 5.30.0 with Bioperl 1.007002 
+-   Puhti: Perl 5.36.0 with Bioperl 1.7.8
```
rkronberg, line 34: This does not seem to be the default version on Puhti. However, the `bioperl/1.007008` module seems to be identical to the `bioperl/1.7.8.` module (is the extra dot there on purpose in the modulefile?) @amsaren

amsaren: Some name confusion due to historical reasons. I think Kimmo M. has at some updated bioperl in module "bioperl/1.007008" but did not change the name. I created a new module with more informative name (points to same environment). I did not want to remove the other module as some users probably have it hard coded in their scripts. I now made made "bioperl/1.7.8" the default version, and also corrected the extra dot (thank for pointing this out).

rkronberg:
I made a bunch more formatting fixes. One thing that I think could be still addressed in this PR is the Trinity page. It is mentioned that the program can create a lot of files (up to 100 000) and that user should apply for more quota because of this. Is this something we want to recommend as a first resort? Or perhaps NVMe and packaging the data would be better? @amsaren

kkmattil:
NVMe is not a solution in the case of trinity as the job execution is typically distributed over several nodes.

rkronberg:
> NVMe is not a solution in the case of trinity as the job execution is typically distributed over several nodes.

Ok I see, thanks for clarifying!

rkronberg (approved):

PR #2212, opened 2024-08-20
Removing the ingress keyword from the load balancer service

vahidmohsseni:
## Proposed changes

The load balancers service needed some modifications.


jemaltahir (approved):
LGTM

PR #2240, opened 2024-09-04
Import CSS custom properties from CSC Design System

joonas-somero:
## Proposed changes

Plug the CSS custom properties from the CSC Design System into the existing mess of sort of corresponding properties. For example, the custom property (also sometimes called a CSS variable) `--csc-success` is now getting its value from the `--c-success-500` property defined in the imported stylesheet:

```diff
--- --csc-success: #51a808;
+++ --csc-success: var(--c-success-500);
```

This way the theme colors are coming from the DS and the existing mess does not have to be cleaned up right now.

It is also now possible to use the [CSC Design System colors](https://design-system.csc.fi/design-tokens/color) in the content:

```markdown
# Page with colors

Some text with a <span style="color: var(--c-secondary-600);">word</span> in the CSC secondary color.
```

For a review, a quick visual check ([Reference Card preview](https://csc-guide-preview.2.rahtiapp.fi/origin/design-system-properties/ref/)) that no element looks broken or out-of-place should be enough. Some colors now have a slightly different shade to conform to the example elements at the [CSC Design System site](https://design-system.csc.fi/).


rkronberg (approved):
Looks good!

PR #2246, opened 2024-09-12
Update the dependencies (and sort out the resulting mess)

joonas-somero:
## Proposed changes

Title.

Preview (developmental deployment) https://csc-guide-preview-devel.2.rahtiapp.fi/origin/update-dependencies/

TODO:
- [x] A _What's new_ entry
- [x] Update dependecies for Preview **(use link above until available at usual address)**
- [x] Announcement (please feel free to edit)


joonas-somero:
I suppose this should wait for Preview to support the new dependencies.

Dockerfile
```diff
@@ -5,32 +5,36 @@ LABEL maintainer="CSC Service Desk <servicedesk@csc.fi>"
...
     git reset HEAD --hard && \
-    git checkout -f $repo_branch
+    git checkout -f $repo_branch; fi
 
-RUN pip3 install --no-cache-dir -r requirements.txt && \
-    bash scripts/generate_alpha.sh && \
+RUN bash scripts/generate_alpha.sh && \
     bash scripts/generate_by_system.sh && \
     bash scripts/generate_new.sh && \
     bash scripts/generate_glossary.sh && \
```
lvarin, line 40: Why not join these two `RUN`s?

If they are joined and the `.git` directory is deleted as a last step, there will be only one layer and it will be smaller.

lvarin:
I checked the changes but not the CSS, because I do not know enough. Maybe @jemaltahir can comment on those changes.

Looks good to me, only the small comment about the `Dockerfile`.

jemaltahir:
I am reviewing the CSS part.

Dockerfile
joonas-somero: I've now joined them. For getting the missing `.git` directory, I was thinking

```dockerfile
RUN if [ ! -d ".git" ]; then \
    git clone --bare https://github.com/$repo_org/$repo_name .git && \
    git init && \
    git switch --force $repo_branch; fi && \
    bash scripts/generate_alpha.sh && \
    bash scripts/generate_by_system.sh && \
    bash scripts/generate_new.sh && \
    bash scripts/generate_glossary.sh && \
    mkdocs build -d /usr/share/nginx/html
```

I don't know if it's any quicker, but it is more readable.

jemaltahir (commented):
LGTM

lvarin (approved):
LGTM

PR #2334, opened 2024-11-14
Updating FEGA pricing document

lkalliok:
Needed to update the document due to changes in policies.


fmorelloCSC (approved):

joonas-somero:
Documents, such as the .pdf file, should be served from the Docs CSC Allas bucket. I can upload it.

PR #2418, opened 2025-02-03
Add image hosting and linking step

samumantha:
As discussed in chat. 

Might require some more linking, and later some more text. 

Feel free to update/edit as you see fit.


tolonenj (approved):
LGTM

PR #2438, opened 2025-02-11
Add mahti NVMe drives and small partition

galfthan:
Add mahti NVMe drives and small partition [https://csc-guide-preview.2.rahtiapp.fi/origin/add-mahti-nvmes/](https://csc-guide-preview.2.rahtiapp.fi/origin/add-mahti-nvmes/)

Additional examples could be good. There are most likely pages that could be amended to reflect this new capability (comparisons of puhti and mahti etc.)


galfthan:
This is now complete from my point of view. Most likely thera are still some pages (applications, tutorials etc.) that could be updated to reflect this change, but this should cover the basics.

rkronberg (approved):

PR #2509, opened 2025-03-25
clarifications to ssh instructions

rkronberg:
https://csc-guide-preview.2.rahtiapp.fi/origin/improve-ssh-keys/computing/connecting/#using-an-ssh-client

Looks like a lot of changes, but basically I just reordered the sections so that it is clearer that the user really needs to first set up the SSH keys before they can try to connect over SSH.

### DIFF
--- a/docs/computing/connecting/ssh-unix.md
+++ b/docs/computing/connecting/ssh-unix.md
@@ -20,15 +20,15 @@ You will be asked to type a passphrase. Please choose a secure passphrase. It
 should be at least 8 characters long and contain numbers, letters and special
 characters. Never leave the passphrase empty!
 
+After you have generated an SSH key pair, you need to add the **public key** to
+the MyCSC portal.
+[Read the instructions here](ssh-keys.md#adding-public-key-in-mycsc).
+
 !!! note "Using SSH keys"
     See the page on [setting up SSH keys](ssh-keys.md) for general
-    information about using SSH keys for authentication.
-
-## Copying public key to supercomputer
-
-Starting April 14 2025, the only way to copy a public key to a supercomputer is
-through the MyCSC customer portal.
-[Read the instructions here](ssh-keys.md#adding-public-key-in-mycsc).
+    information about using SSH keys for authentication. Note that copying the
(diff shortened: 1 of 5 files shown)


galfthan (approved):
I agree with the changes

docs/computing/connecting/ssh-unix.md
```diff
@@ -3,12 +3,36 @@
 --8<-- "auth-update-ssh.md"
 
 On Unix-based systems like macOS and Linux, it is recommended to connect to CSC
-supercomputers using the pre-installed terminal program.
+supercomputers using the pre-installed terminal program. The OpenSSH client
+typically comes pre-installed on macOS and Linux systems.
+
+## Generating SSH keys
+
+On macOS and Linux, you can use the `ssh-keygen` command-line utility for
```
galfthan, line 11: Reading the page this section is now not really clear for somebody who does not know what ssh keys are. Why do we need them? Should we add a sentance or two that says that to access supercomputers you need to set up ssh keys. 

Perhaps merge this section and the next on copying to supercomputers?

docs/computing/connecting/ssh-windows.md
```diff
@@ -9,16 +9,36 @@ popular alternatives: [PowerShell](#powershell), [PuTTY](#putty) and
...
 command-line shell to connect to a CSC supercomputer using the
 [Win32 OpenSSH client](https://learn.microsoft.com/en-us/windows-server/administration/openssh/openssh_install_firstuse).
 To install OpenSSH on a Windows device, follow
 [these installation instructions](https://learn.microsoft.com/en-us/windows-server/administration/openssh/openssh_install_firstuse?tabs=gui#install-openssh-for-windows).
-After installing OpenSSH, you can connect to a CSC supercomputer by opening
-PowerShell and running:
+
+### Generating SSH keys (PowerShell)
+
+After installing OpenSSH, you can generate SSH keys using PowerShell by
```
galfthan, line 21: perhaps here too a sentance explaining why?

galfthan (changes requested):
New review after reading through preview again

joonas-somero (approved):

PR #2510, opened 2025-03-25
Decrease expiration time in filesender CLI example

rkronberg:
There were some tickets about too long expiration times causing errors. Decreasing to 7 days should fix these.


attesillanpaa (approved):
ok

PR #2524, opened 2025-04-01
Info box about MFA coming to web interfaces

rkronberg:
https://csc-guide-preview.2.rahtiapp.fi/origin/ood-mfa/computing/webinterface/connecting/


galfthan (changes requested):
The instructions should emphasize that home organization haka is preferred and csc mfa should only be used when that is not available

PR #2624, opened 2025-06-26
FAQ for BU renewal

rkronberg:
## Proposed changes

https://csc-guide-preview.2.rahtiapp.fi/origin/bu-renewal-faq/support/faq/bu-renewal/

- [x] I have followed the instructions in the [Contributing](https://github.com/CSCfi/csc-user-guide/blob/master/CONTRIBUTING.md) and [Styleguide](https://github.com/CSCfi/csc-user-guide/blob/master/STYLEGUIDE.md) documents.
- [x] My pull request passes the automated tests. If not, visit [pull requests at Travis CI](https://app.travis-ci.com/github/CSCfi/csc-user-guide/pull_requests) and have a look at the job log.
- [x] I have included a link to the appropriate [preview page](https://csc-guide-preview.2.rahtiapp.fi/origin/) (select your branch from the list).

### DIFF
--- a/docs/support/faq/bu-renewal.md
+++ b/docs/support/faq/bu-renewal.md
@@ -62,7 +62,7 @@ based on the type of billing unit applied for.
 The billing unit renewal introduces a "cutter" that will depreciate unused BUs.
 The billing unit cutter only affects projects that have used **less than 40% of their
 billing units** in a 6 month period counted from the last billing unit grant
-or deprecation check. If a project has used more than that, no resources are
+or depreciation check. If a project has used more than that, no resources are
 cut. The billing unit cutter affects only academic projects with student
 projects being exempt.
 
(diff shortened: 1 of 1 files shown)


attesillanpaa:
Looks good! Maybe add a calming note along the lines "do I need to do something when the change comes?" -> no, your BUs will be converted automatically, and you can continue working as usual. Maybe also a link to myCSC to check your current BU levels? (or is this getting outside the scope?)

nurmijoo:
These look good! I can think of a few more that could be added:
- "How will the BU renewal affect CSC's services?" &rarr; The migration to new Billing units will have minimal effect on service usage and services will be available as usual. During the migration resource applications are closed however.
- "Will I lose Billing units to the new BU cutter even if my projects use billing units?" &rarr; The billing unit cutter only affect projects that use less than 40% of their Billing units in the 6 month period, counted from the last Billing unit grant or deprecation check. If a project has used more than that, no resources are cut. The Billing unit cutter affects only academic projects with student projects being exempt.
- "When will the Billing unit deprecation happen?"  &rarr; The Billing unit cutter is taken into use at the same time as the new units, so the first cuts would happen in March 2026 at the earliest. The Billing unit usage of projects is checked six months after either the latest Billing unit grant or the latest check.

And I agree on the "will I need to do something?" question, it is a good way of reassuring users that they need not panic about this

attesillanpaa (approved):

PR #2769, opened 2025-11-03
Add reference to OOD File Deletion Explorer clean-up-data.md

jlento:
This reference (in the lcleaner section) can (and should) be published when the tool is available after OOD update, but before the the file cleaning email is sent to customers 5.11.2025.

https://csc-guide-preview.2.rahtiapp.fi/origin/file-deletion-explorer

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -1,0 +1,26 @@
+// These assume that Conda is installed in the default location
+// and that 'docs-env' has the requirements installed.
+{
+  "version": "0.2.0",
+  "configurations": [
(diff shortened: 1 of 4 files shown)


CSC-swesters:
Thanks Juha! I pushed a few corrections, let me know if they are okay.

I was wondering if the File deletion explorer app will have its own documentation somewhere, does @robinkar have plans for that? Maybe we should be linking to that instead of merging this separate documentation?

jlento (commented):
Looks good!

PR #2822, opened 2025-12-31
Improve Python venv documentation

Iidahak:
Add instructions on how to pip install packages that require compilation. 
Clarify that venv needs to be created with the same python module as the one it is used with.


Iidahak:
Warning-box fixed.

PR #2842, opened 2026-01-20
Tykky env feature and other improvements

Traubert:
## Proposed changes

[Preview page here](https://csc-guide-preview.2.rahtiapp.fi/origin/tykky-env-feature-and-other-improvements/computing/containers/tykky/).

The main purpose is to document the new environment feature in tykky, but while I was at it, I did some additional work on it. Hopefully the main idea of tykky and using it is clearer now. 

- [x] I have followed the instructions in the [Contributing](https://github.com/CSCfi/csc-user-guide/blob/master/CONTRIBUTING.md) and [Styleguide](https://github.com/CSCfi/csc-user-guide/blob/master/STYLEGUIDE.md) documents.
- [x] My pull request passes the automated tests. If not, visit [pull requests at Travis CI](https://app.travis-ci.com/github/CSCfi/csc-user-guide/pull_requests) and have a look at the job log.
- [x] I have included a link to the appropriate [preview page](https://csc-guide-preview.2.rahtiapp.fi/origin/) (select your branch from the list).

### DIFF
--- a/docs/computing/containers/tykky.md
+++ b/docs/computing/containers/tykky.md
@@ -60,11 +60,10 @@ will be run by default, instead of the normal system-provided `python3`. You can
 file](../modules.md) do this for you, and in fact that is how many of the modules in the CSC
 environment have been produced.
 
-If you would like a less permanent containerized environment, tykky has a "venv-like" environmant
-feature. You can activate a tykky installation with `tykky activate <install_dir>`, where
-`<install_dir>` is the directory tykky used during installation. Your prompt will then show
-`(install_dir)`, and the installed executables will be automatically used. You can get back to
-your previous environment with `tykky deactivate`.
+Instead of manually modifying `$PATH`, you can also activate a tykky installation with `tykky
+activate <install_dir>`, where `<install_dir>` is the target directory tykky used during
+installation. Your prompt will then show `(install_dir)`, and the installed executables will be
+automatically used. You can get back to your previous environment with `tykky deactivate`.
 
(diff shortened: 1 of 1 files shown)


docs/computing/containers/tykky.md
```diff
@@ -44,6 +41,26 @@ module purge
...
 ```
 
+## Overview
+
+Tykky provides commands ending in `-containerize` (`pip-`, `conda-`) and starting with `wrap-`
+(`-container`, `-install`). They all produce a container, and place in a given target directory a
+`bin/` subdirectory, which contains executables for you to run. For example, if you are making an
+environment with Python libraries, `bin/python3` would behave like a Python with those dependencies
+installed. It will run in a container, but thanks to tykky, it will be able to see the file system
+without you needing to think about it.
```
mvsjober:
```suggestion
Tykky provides the commands `conda-containerize` and `pip-containerize` which produce a
containerized environment based on a Conda environment file or a pip requirements file
respectively. The command `wrap-install` wraps an existing Conda installation into a container. The
command `wrap-container` takes an existing container and adds wrapper scripts to make it work in the
same way as other tykky-based environments.

In all cases, a containerized environment is produced into a given target directory. That directory
will contain a `bin/` subdirectory with executables to run. For example, if you are making an
environment with Python libraries, `bin/python3` would behave like a Python with the given
dependencies installed. It will run in a container, but thanks to tykky, it will work transparently
as if it was running directly on the normal system.
```

PR #2884, opened 2026-02-26
Add Fairdata Services to MFA listing

csckata:
Fairdata Services are taking MFA into use in March, so it would be appropriate to add it to the listing. If needed, more information can be asked from me or Erja Kortelainen (Fairdata PO).


DeRuina (approved):
LGTM

PR #2915, opened 2026-03-27
WIP Update Python packages

joonas-somero:
# Updates for packages in requirements.txt

- [Preview](https://csc-guide-preview.2.rahtiapp.fi/origin/update-python-requirements/)

This was originally about [some Dependabot alerts](https://github.com/CSCfi/csc-user-guide/security/dependabot) (the last one, about Pygments, doesn't have a fix yet, BTW).

It turns out that a warning is now logged when running `mkdocs build`. Setting the environment variable they mention only removes the WARNING type from logger while the message above the hyphens is still logged. The mentioned drop-in replacement for MkDocs, ProperDocs, was included as a dependency for Material for MkDocs, so at the time of writing there are two redundant "MkDocses" ("MkDocii?") in requirements.txt.

```text

 │  ⚠  Warning from the Material for MkDocs team
 │
 │  MkDocs 2.0, the underlying framework of Material for MkDocs,
 │  will introduce backward-incompatible changes, including:
 │
 │  × All plugins will stop working – the plugin system has been removed
 │  × All theme overrides will break – the theming system has been rewritten
 │  × No migration path exists – existing projects cannot be upgraded
 │  × Closed contribution model – community members can't report bugs
 │  × Currently unlicensed – unsuitable for production use
 │
 │  Our full analysis:
 │
 │  https://squidfunk.github.io/mkdocs-material/blog/2026/02/18/mkdocs-2.0/

----------------------------------------------------------------

WARNING: MkDocs may break support for all existing plugins and themes soon!

The owner of MkDocs has completely abandoned maintenance of the project, and instead is planning to publish a "version 2" which will not support any existing themes, plugins or even your configuration files. This v2 may eventually replace what you download with `pip install mkdocs`, suddenly breaking the build of your existing site.

To avoid these risks, switch to ProperDocs, a continuation of MkDocs 1.x and a drop-in replacement that supports your current MkDocs setup.
Simply install it with `pip install properdocs` and build your site with `properdocs build` instead of the MkDocs equivalents.

Alternatively, to just skip this warning in the future, you can set the environment variable `DISABLE_MKDOCS_2_WARNING=true`.

For more info visit https://github.com/ProperDocs/properdocs/discussions/33 and https://properdocs.org/

(This warning was initiated by one of the plugins that you depend on.)

----------------------------------------------------------------
```


joonas-somero:
We'll go with ProperDocs and MaterialX.

PR #2989, opened 2026-05-06
Roihu user spack tough case

jlento:
What to do when spack is not using dependencies from the upstream

https://csc-guide-preview.2.rahtiapp.fi/origin/roihu-useer-spack-tough-case/support/tutorials/roihu-user-spack


docs/support/tutorials/roihu-user-spack.md
aantao, line 56: The Spack variables should be exported before activating spack.

jlento: > The Spack variables should be exported before activating spack.

Swapped the order

PR #3166, opened 2026-08-11
Update cdo.md

jlento:
https://csc-guide-preview.2.rahtiapp.fi/origin/cdo-version-and-standalone/apps/cdo/


docs/apps/cdo.md
```diff
@@ -24,14 +24,10 @@ CDO is free to use and open source under the GNU General Public License, version
 
 CDO is available with following versions:
 
+* 2.6.0 - standalone executable with `module use cdo` command
```
aantao: `module use cdo` doesn't work for me. I think it should be `module load cdo` .

jlento: Fixed!

aantao (approved):
Looks good to me!

PR #3204, opened 2026-08-26
Update mkdocs-section-index plugin

joonas-somero:
## Proposed changes

Update mkdocs-section-index plugin, since the index sections were not working after #3187 last week.

No preview.


rkronberg (approved):
ok

